PR #2, opened 2026-05-18
[core] Add column statistics and tests

JingsongLi:
- stats.rs: per-column min/max/null-count statistics with serialization/deserialization support
- Wire stats into writer (collection + serialization) and reader (deserialization + row_group_stats API)
- reader_tests.rs: unit tests for reader functionality
- robustness_test.rs: edge-case and corruption tests
- stress_test.rs: large-scale and performance tests

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -21,6 +21,7 @@ pub mod bucket_writer;
 pub mod reader;
 pub mod schema;
 pub mod spec;
+pub mod stats;
 pub mod types;
 pub mod values;
 pub mod varint;
--- a/core/src/reader.rs
+++ b/core/src/reader.rs
@@ -24,6 +24,7 @@ use arrow_schema::{DataType, Field, Schema};
 use crate::bucket_reader::{read_typed_value, read_variable_value, BucketReader, ColumnPageReader};
 use crate::schema::MosaicSchema;
 use crate::spec::*;
+use crate::stats::{self, ColumnStats};
 use crate::types;
 use crate::values::Value;
 use crate::varint;
@@ -122,6 +123,7 @@ pub struct RowGroupMeta {
     pub num_rows: usize,
     pub bucket_offsets: Vec<u64>,
     pub bucket_layouts: Vec<BucketLayout>,
+    pub stats: Vec<ColumnStats>,
 }
 
 pub trait ReaderAccess {
@@ -133,6 +135,7 @@ pub trait ReaderAccess {
         rg_index: usize,
         columns: &[usize],
     ) -> io::Result<RowGroupReader>;
+    fn row_group_stats(&self, rg_index: usize) -> io::Result<&[ColumnStats]>;
 }
 
 pub struct MosaicReader<I: InputFile> {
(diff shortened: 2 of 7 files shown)


leaves12138 (changes requested):
Thanks for adding the reader coverage and row-group statistics. The overall direction looks good, but I found one correctness blocker and one CI/test-suite concern that should be addressed before merging.

1. **Blocking: enabling stats on large `Decimal128` columns can make the file unreadable.**

   `supports_stats` currently accepts every `DataType::Decimal128(_, _)`, but `extract_value_for_stats` only returns a value when `precision <= 18`; for larger decimal precision it returns `None`. `serialize_stats` then writes only `column_index` and `null_count` when `min/max` are missing, while `deserialize_stats` assumes min/max bytes are present whenever `null_count < num_rows`. This corrupts the row-group index stream.

   I reproduced this locally with a temporary test using `Decimal128(30, 5)` and `WriterOptions { stats_columns: vec![0], ... }`; `MosaicReader::new` failed with `varint: unexpected end of data` after writing two non-null decimal rows.

   Please either restrict stats support to decimal types that can actually produce and serialize min/max values, e.g. `Decimal128(p, _) if p <= 18`, or add an explicit per-column stats presence flag plus proper `DecimalLarge` serialization/deserialization. It would also be safer for stats serialization to return an error or skip the stat entry if `null_count < num_rows` but min/max are absent.

2. **Please keep large stress tests out of the default `cargo test` path.**

   The new CI runs plain `cargo test`, so tests such as `core/tests/stress_test.rs::test_200mb_mixed_types` are executed on every PR. That test builds 5,000,000 rows in debug mode and did not finish within a 25-second local timeout when run alone; `test_paged_bucket_large` is also fairly large. These are useful stress/benchmark cases, but they should be scaled down for normal CI or marked `#[ignore]` / gated behind a feature so the default PR validation remains fast and stable.

Validation I ran locally on head `c716e84`:

- `cargo fmt --all -- --check`: passed.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed.
- Existing unit tests and robustness tests passed before the large stress tests.
- A targeted temporary repro for `Decimal128(30, 5)` stats failed as described above.

leaves12138 (approved):
Thanks for the update. I reviewed the latest head (`b653ee9`). The previous `Decimal128` stats blocker is fixed by restricting statistics support to `Decimal128(0..=18, _)`, so large-precision decimals are no longer selected for row-group stats.

Validation I ran locally:

- Temporary repro for `Decimal128(30, 5)` with `stats_columns: vec![0]`: passed; the file opens successfully and stats are skipped for that column.
- `cargo fmt --all -- --check`: passed.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed.
- `cargo test --workspace --all-targets`: passed (95 unit tests, 53 robustness tests, 21 stress tests). The full test run completed locally, including `test_200mb_mixed_types`.

GitHub Actions for the latest head are still queued from my side, but local validation is clean and I do not see any remaining blocker. Non-blocking: adding a small committed regression test for `Decimal128` precision > 18 with stats enabled would make this fix harder to regress later.

PR #3, opened 2026-05-18
fix: fix github ci

JingsongLi:


leaves12138 (approved):
Thanks for the update. I reviewed the latest head (`c82bab1`) and verified both the workflow change and the Rust-side clippy fixes.

Validation results:

- GitHub Actions for this head are green: `check` and `build-and-test` both passed.
- `cargo fmt --all -- --check`: passed locally.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed locally.
- `cargo test --workspace --all-targets`: passed locally (22 tests).

The CI workflow now covers formatting, clippy, build, and tests, and the code changes are behavior-preserving cleanup for the current logic. I do not see any blocker for this PR.

PR #4, opened 2026-05-18
Add ffi and cpp

JingsongLi:


leaves12138 (changes requested):
Thanks for adding the Rust FFI and C++ wrapper. The Rust side looks healthy, but I found a C/C++ API build blocker that should be fixed before merging.

**Blocking: the generated `include/mosaic.h` is not self-contained and exposes Rust-only Arrow FFI type names.**

After running the Rust build, `ffi/build.rs` generates `include/mosaic.h` with signatures such as:

```c
struct MosaicWriterHandle *mosaic_writer_open(..., FFI_ArrowSchema *ffi_schema, ...);
int mosaic_writer_write_batch(..., FFI_ArrowArray *ffi_array, FFI_ArrowSchema *ffi_schema);
```

However, `FFI_ArrowSchema` and `FFI_ArrowArray` are Rust crate type names, not the C Arrow Data Interface names. The public C/C++ API should expose `struct ArrowSchema` and `struct ArrowArray` (and include/require the Arrow C Data Interface header), otherwise downstream C/C++ users cannot compile against the generated header.

I reproduced this with a minimal header smoke test after `cargo clippy` generated `include/mosaic.h`:

```bash
gcc -I. -fsyntax-only /tmp/mosaic_h_smoke.c
```

where `/tmp/mosaic_h_smoke.c` only contains:

```c
#include "include/mosaic.h"
int main(void) { return 0; }
```

It fails with unknown type errors for `FFI_ArrowSchema` and `FFI_ArrowArray`. `include/mosaic.hpp` also includes `mosaic.h` before any Arrow C ABI declaration, so the C++ wrapper has the same issue.

Please make the generated C API use the real Arrow C Data Interface types, e.g. expose `struct ArrowSchema` / `struct ArrowArray` in the generated header and include the proper Arrow C ABI header (or otherwise make the wrapper include it before `mosaic.h`).

One additional FFI-safety note: the C++ callback adapters in `include/mosaic.hpp` call `std::function` directly from Rust-invoked `extern "C"` callbacks. If a user callback throws, the C++ exception can unwind through Rust frames, which is undefined behavior. It would be safer to mark those adapters `noexcept`, catch exceptions, and return an error code.

Validation I ran locally:
- `cargo fmt --all -- --check`: passed.
- `cargo clippy --all-targets --workspace -- -D warnings`: passed.
- `cargo test`: passed.
- Minimal `include/mosaic.h` C syntax smoke test: failed as described above.

leaves12138 (changes requested):
Thanks for the update. The previous `FFI_ArrowSchema` / `FFI_ArrowArray` issue is fixed after regenerating `include/mosaic.h`: it now includes `arrow_c_data.h` and exposes `ArrowSchema` / `ArrowArray`. The callback adapters are also now `noexcept` and catch exceptions. I reran the Rust checks successfully.

I found one remaining C++ integration blocker:

**Blocking: the vendored `arrow_c_data.h` conflicts with Arrow C++'s official `arrow/c/abi.h`.**

`include/arrow_c_data.h` defines `ArrowSchema` and `ArrowArray`, but it does not use the official Arrow C Data Interface guard macro (`ARROW_C_DATA_INTERFACE`). This means any C++ translation unit that includes Mosaic and then Arrow's C++ bridge headers will see the structs defined twice.

This is exactly what the PR's own C++ test does:

```cpp
#include "mosaic.hpp"
#include <arrow/api.h>
#include <arrow/c/bridge.h>
```

I reproduced the same issue locally by compiling a minimal TU that includes `mosaic.hpp` and then Arrow's official `arrow/c/abi.h`:

```text
error: redefinition of 'struct ArrowSchema'
error: redefinition of 'struct ArrowArray'
```

Please make `include/arrow_c_data.h` compatible with the official Arrow C Data Interface header. For example, guard the struct definitions with `#ifndef ARROW_C_DATA_INTERFACE` / `#define ARROW_C_DATA_INTERFACE`, and keep `typedef struct ArrowSchema ArrowSchema;` / `typedef struct ArrowArray ArrowArray;` aliases outside that block so the generated C header can still use `ArrowSchema *` in C. Alternatively, include/use Arrow's official C ABI header directly instead of vendoring a conflicting copy.

Validation I ran on head `70bc6dd`:
- `cargo build -p mosaic-ffi`: passed and regenerated `include/mosaic.h` with `ArrowSchema` / `ArrowArray`.
- Minimal `#include "mosaic.h"` C syntax smoke test: passed.
- Minimal `#include "mosaic.hpp"` C++ syntax smoke test: passed.
- `cargo fmt --all -- --check`: passed.
- `cargo clippy --all-targets --workspace -- -D warnings`: passed.
- `cargo test`: passed.
- `#include "mosaic.hpp"` followed by Arrow's official `arrow/c/abi.h`: fails with the redefinition above.

leaves12138 (approved):
Thanks for the update. The previous blockers are fixed in `cd2014d`.

I revalidated the FFI/C++ header boundary locally:
- `cargo build -p mosaic-ffi` regenerates `include/mosaic.h` with `ArrowSchema` / `ArrowArray` and `#include "arrow_c_data.h"`.
- Minimal `#include "mosaic.h"` C syntax smoke test passes.
- Minimal `#include "mosaic.hpp"` C++ syntax smoke test passes.
- Both include orders pass with Arrow's official C ABI header: `mosaic.hpp` then `arrow/c/abi.h`, and `arrow/c/abi.h` then `mosaic.hpp`.
- `cargo fmt --all -- --check` passes.
- `cargo clippy --all-targets --workspace -- -D warnings` passes.
- `cargo test` passes.

I did not run the full linked C++ Arrow test locally because this machine does not have Arrow C++ installed; the new CI job should cover that part. The code looks good to me pending CI.

PR #5, opened 2026-05-18
feat: optimize projected read buffer copies

QuakeWang:
Projected reads already avoided unnecessary IO for sparse paged buckets, but the reader still copied buffers at several internal boundaries.

Before this PR:
  - `read_ranges` coalesced IO ranges, then copied each logical range into its own `Vec<u8>`.
  - Paged projection copied selected slot bytes into an intermediate map before parsing.
  - `parse_column_slot` copied the decompressed page payload after parsing its header.

This PR keeps the existing format and `read_ranges` API, but adds a shared range buffer path for projected reads. The reader now borrows slices from coalesced buffers during parsing, stores paged slot locations instead of slot bytes, and lets   `ColumnPageReader` parse from an offset inside the owned decompressed page buffer.

### DIFF
--- a/core/src/bucket_reader.rs
+++ b/core/src/bucket_reader.rs
@@ -565,6 +565,33 @@ impl ColumnPageReader {
         page_data: Vec<u8>,
         num_rows: usize,
     ) -> io::Result<Self> {
+        Self::new_with_page_data_start(
+            col_type,
+            encoding,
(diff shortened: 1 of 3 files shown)


JingsongLi:
Nice Catch!

JingsongLi (approved):
+1

PR #6, opened 2026-05-19
Add Java/JNI bindings and CI test job

JingsongLi:
Enabling Java applications to use MosaicWriter/MosaicReader via Arrow C Data Interface over JNI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,3 +102,32 @@ jobs:
         run: cpp/build/test_mosaic
         env:
           LD_LIBRARY_PATH: ${{ github.workspace }}/target/release
+
+  java-test:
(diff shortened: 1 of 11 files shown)


leaves12138 (approved):
Looks good to me.

I reviewed the Java/JNI binding path, including the Arrow C Data import/export flow, native handle lifecycle, Java `InputFile`/`OutputStream` bridging, row-group projection/statistics APIs, and the new CI job. The JNI library builds successfully locally, `cargo test --workspace` passes, and the GitHub `java-test`, `build-and-test`, `cpp-test`, and `check` jobs are all green for this head commit. The local Java test run was not applicable in my environment because it is still on JDK 8 while the PR correctly configures CI with JDK 11 for Arrow 18.

PR #7, opened 2026-05-19
fix: preserve paged all-null projections

QuakeWang:
## Problem

Paged buckets store `ALL_NULL` columns as directory entries with `slot_size = 0` and no slot payload. The reader correctly skipped round 2 IO for those slots, but partial projection also used the presence of round 2 ranges to decide whether to construct paged bucket readers.

As a result, when a projection requested only all-null columns from a partial paged bucket, the requested column could be dropped from the output.

## Fix

Separate the two concepts:

- `slot_size > 0` decides whether a projected column needs round 2 IO.
- `projected[gi]` decides whether a column needs to appear in the result.

The reader now tracks partial paged buckets explicitly and constructs their readers even when no round 2 IO is required. `ALL_NULL` projected slots are materialized with an `ENCODING_ALL_NULL` reader.

## Tests

- Added a regression test for projecting only an all-null paged column.
- Ran `cargo test --workspace`.
- Ran `cargo clippy --workspace --all-targets -- -D warnings`.

### DIFF
--- a/core/src/reader.rs
+++ b/core/src/reader.rs
@@ -599,6 +599,7 @@ impl<I: InputFile> ReaderAccess for MosaicReader<I> {
 
         // Per-bucket directory parse results (slot_sizes, slot_file_offsets) for paged buckets
         let mut paged_dir_info: Vec<Option<(Vec<usize>, Vec<u64>)>> = vec![None; self.num_buckets];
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #8, opened 2026-05-19
Add Python bindings and CI test job

JingsongLi:


python/mosaic/mosaic.py
```diff
@@ -0,0 +1,367 @@
...
+                _check_error("record_batch_export failed")
+            return pa.RecordBatch._import_from_c(array_ptr, schema_ptr)
+        finally:
+            lib.mosaic_record_batch_free(rb_handle)
+
+    def read_all(self, columns=None):
+        batches = []
+        for rg in range(self.num_row_groups):
+            batches.append(self.read_row_group(rg, columns=columns))
+        return pa.Table.from_batches(batches, schema=batches[0].schema) if batches else pa.table({})
```
leaves12138: `read_all()` currently returns `pa.table({})` when the file has no row groups, which drops the Mosaic file schema. This can happen for a valid empty file (for example writing a zero-row `RecordBatch`), and `test_zero_rows` only checks `num_rows`, so it misses the schema loss.

Please preserve the reader schema in the empty case, e.g. return an empty table with `self.schema` (and the projected schema if `columns` is provided), and extend the zero-row test to assert the schema/field names as well.

leaves12138 (changes requested):
Thanks for adding the Python bindings. The overall FFI shape looks good, but I found one correctness issue in the empty-table path that should be fixed before merging.

I reviewed the Python ctypes declarations, Arrow C Data import/export ownership, callback lifetimes, reader/writer handle lifecycle, and the new CI job. I also verified `cargo build --release -p mosaic-ffi` and `cargo test --workspace` locally. The local Python test run was not applicable in my environment because the default Python is 3.6 and cannot install a compatible `pyarrow` wheel, and GitHub CI was still queued at the time of review.

python/mosaic/mosaic.py
```diff
@@ -0,0 +1,372 @@
...
+    def read_all(self, columns=None):
+        batches = []
+        for rg in range(self.num_row_groups):
+            batches.append(self.read_row_group(rg, columns=columns))
+        if batches:
+            return pa.Table.from_batches(batches, schema=batches[0].schema)
+        if columns is not None:
+            projected_schema = pa.schema([self._schema.field(i) for i in columns])
+            return pa.table({f.name: pa.array([], type=f.type) for f in projected_schema})
+        return pa.table({f.name: pa.array([], type=f.type) for f in self._schema})
```
leaves12138: This still does not fully preserve the file schema for empty results. Building `pa.table({...})` from arrays re-infers fields and loses field-level metadata such as nullability. For example, an empty file with `pa.field("id", pa.int32(), nullable=False)` is read back as nullable here.

Please return an empty table with the explicit schema instead, e.g. select the projected schema when `columns` is provided and use `pa.Table.from_batches([], schema=schema)` (or another explicit-schema construction). The zero-row test should also assert schema equality / nullability, not only names and types.

leaves12138 (changes requested):
Thanks for the quick update. The empty result now preserves column names and types, and the existing Python test suite passes locally with Python 3.10. However, the empty-table path still does not preserve the full Arrow schema, so I think this needs one more small fix.

I ran:

- `cargo build --release -p mosaic-ffi`
- `python -m pytest -q tests/test_mosaic.py` with Python 3.10.14 / pyarrow 24.0.0

I also checked an empty table with a non-nullable field; `reader.schema` remains `id: int32 not null`, but `read_all().schema` becomes `id: int32` (nullable).

leaves12138 (approved):
Looks good to me now.

The latest update fixes the empty-table path by constructing the empty result with an explicit schema, so field-level properties such as nullability are preserved. The zero-row test now covers a non-nullable field as well.

I validated the latest files from head commit `2e0b54a7ffec6b690c8e698a8d1a4a904060571e` locally with Python 3.10.14 / pyarrow 24.0.0:

- `cargo build --release -p mosaic-ffi`
- `python -m pytest -q tests/test_mosaic.py`
- an additional empty-table/projection check that verifies `read_all()` preserves the full schema, including `not null`

All passed. GitHub CI was still queued at the time of this review.

PR #9, opened 2026-05-19
Add docs: design, rust-api, cpp-api, java-api and python-api

JingsongLi:


docs/design.html
```diff
@@ -0,0 +1,781 @@
...
+varint   numRules
+repeated numRules times:
+    1 byte   left               (left token of merge rule)
+    1 byte   right              (right token of merge rule)
+
+--- per column (repeated numColumns times, name-sorted order) ---
+varint   sharedPrefixLen       (bytes shared with previous column name)
+varint   suffixLen             (bytes of new suffix)
+bytes    suffix                (suffixLen bytes, raw or BPE-encoded)
+TypeDescriptor</code></pre>
```
leaves12138, line 696: This schema layout is missing the serialized logical/global column index. The implementation writes `global_idx` between `suffix` and `TypeDescriptor` (`core/src/schema.rs`), and the reader decodes it before `deserialize_field`. Without documenting that varint, an independent reader/writer based on this spec would be off by one field and could not preserve the original schema column order.

docs/rust-api.html
```diff
@@ -0,0 +1,254 @@
...
+
+            <h3>Writing with Stats</h3>
+<pre><code><span class="kw">let</span> <span class="kw">mut</span> writer = MosaicWriter::new(output, &amp;arrow_schema, WriterOptions {
+    stats_columns: <span class="kw">vec!</span>[<span class="num">0</span>, <span class="num">2</span>],  <span class="cmt">// build stats for columns 0 and 2</span>
+    ..Default::default()
+})?;</code></pre>
+
+            <h3>Reading Stats</h3>
+<pre><code><span class="kw">for</span> rg_idx <span class="kw">in</span> <span class="num">0</span>..reader.num_row_groups() {
+    <span class="kw">let</span> stats = reader.row_group_stats(rg_idx);
```
leaves12138: `row_group_stats` returns `io::Result<&[ColumnStats]>` in `ReaderAccess`, so this snippet does not compile as written. It should unwrap/propagate the result before iterating, e.g. `let stats = reader.row_group_stats(rg_idx)?;` or `.unwrap()` in a sample.

docs/cpp-api.html
```diff
@@ -0,0 +1,399 @@
...
+        <span class="kw">int64_t</span> pos = <span class="num">0</span>;
+
+        <span class="ty">mosaic</span>::<span class="ty">OutputFile</span> cbs;
+        cbs.write_fn = [file, pos](<span class="kw">const uint8_t</span>* data, <span class="kw">size_t</span> len) <span class="kw">mutable</span> -&gt; <span class="kw">int</span> {
+            <span class="kw">size_t</span> written = std::fwrite(data, <span class="num">1</span>, len, file.get());
+            pos += <span class="kw">static_cast</span>&lt;<span class="kw">int64_t</span>&gt;(written);
+            <span class="kw">return</span> (written == len) ? <span class="num">0</span> : <span class="num">-1</span>;
+        };
+        cbs.flush_fn = [file]() -&gt; <span class="kw">int</span> { <span class="kw">return</span> std::fflush(file.get()); };
+        cbs.get_pos_fn = [file, pos]() <span class="kw">mutable</span> -&gt; <span class="kw">int64_t</span> { <span class="kw">return</span> pos; };
```
leaves12138: The two lambdas capture separate copies of `pos`, so `get_pos_fn` always returns its own unchanged copy (0). The writer relies on `get_pos_fn` for file offsets, so this example can produce an invalid file. The tested code uses shared buffer state (`buf.pos`) for both callbacks; this snippet should do the same or use `ftell`/shared state.

docs/cpp-api.html
```diff
@@ -0,0 +1,399 @@
...
+<span class="cmt">// batch contains only the projected columns</span></code></pre>
+
+            <h3>Column Statistics (Filter Pushdown)</h3>
+            <p>
+                When stats columns are configured during writing, the reader can access per-row-group
+                min/max statistics to skip row groups that don't match a filter predicate:
+            </p>
+<pre><code><span class="cmt">// Writing with stats (arrow_schema is an ArrowSchema* from C Data Interface)</span>
+<span class="ty">mosaic</span>::<span class="ty">Writer</span> writer(std::move(cbs), arrow_schema, {
+    .compression = <span class="num">1</span>,
```
leaves12138: This says "Writing with stats" but it only sets `compression`; `stats_columns` and `num_stats_columns` remain unset, so `get_row_group_statistics` will return an empty vector. Please include a `uint32_t stats_cols[] = { ... }` and set both `opts.stats_columns` and `opts.num_stats_columns`.

docs/java-api.html
```diff
@@ -0,0 +1,386 @@
...
+            <h2>Writing a File</h2>
+
+            <h3>1. Define an Arrow Schema</h3>
+<pre><code><span class="kw">import</span> org.apache.arrow.vector.types.pojo.*;
+<span class="kw">import</span> org.apache.arrow.vector.types.*;
+
+<span class="ty">Schema</span> arrowSchema = <span class="kw">new</span> <span class="ty">Schema</span>(<span class="ty">Arrays</span>.asList(
+    <span class="ty">Field</span>.notNullable(<span class="str">"id"</span>, <span class="kw">new</span> <span class="ty">ArrowType.Int</span>(<span class="num">32</span>, <span class="kw">true</span>)),
+    <span class="ty">Field</span>.nullable(<span class="str">"name"</span>, <span class="ty">ArrowType.Utf8</span>.INSTANCE),
+    <span class="ty">Field</span>.nullable(<span class="str">"score"</span>, <span class="kw">new</span> <span class="ty">ArrowType.FloatingPoint</span>(DOUBLE)),
```
leaves12138: The Java snippets use `new ArrowType.FloatingPoint(DOUBLE)`, but `DOUBLE` is not in scope with the shown imports. The implementation/tests use `FloatingPointPrecision.DOUBLE` (or a static import would be required). Please update all Java snippets that use `DOUBLE` so users can copy/paste them.

leaves12138 (changes requested):
Thanks for adding the docs. I focused on whether the documentation matches the current implementation and found several mismatches that should be fixed before merge. Local static link validation passed; the issues below are about API/format accuracy.

docs/design.html
```diff
@@ -0,0 +1,781 @@
...
+                    <tr><td>8</td><td>CHAR</td><td>varint length</td></tr>
+                    <tr><td>9</td><td>VARCHAR</td><td>varint length</td></tr>
+                    <tr><td>10</td><td>STRING</td><td>(none) &mdash; VARCHAR with MAX_LENGTH</td></tr>
+                    <tr><td>11</td><td>BINARY</td><td>varint length</td></tr>
+                    <tr><td>12</td><td>VARBINARY</td><td>varint length</td></tr>
+                    <tr><td>13</td><td>BYTES</td><td>(none) &mdash; VARBINARY with MAX_LENGTH</td></tr>
+                    <tr><td>14</td><td>DECIMAL</td><td>varint precision, varint scale</td></tr>
+                    <tr><td>15</td><td>TIME</td><td>varint precision</td></tr>
+                    <tr><td>16</td><td>TIMESTAMP</td><td>varint precision</td></tr>
+                    <tr><td>17</td><td>TIMESTAMP_LTZ</td><td>varint precision</td></tr>
```
leaves12138: For TIMESTAMP_LTZ/Timestamp-with-timezone, the implementation writes more than just `precision`: it also writes the timezone string length and bytes (`core/src/types.rs`). The spec should include `varint timezoneLength` and `bytes timezone`, otherwise the documented type descriptor does not match files written by the current code.

docs/cpp-api.html
```diff
@@ -0,0 +1,400 @@
...
+             age_builder.Finish().ValueOrDie(),
+             score_builder.Finish().ValueOrDie()});
+
+        <span class="cmt">// 3. Export via Arrow C Data Interface, create writer, and write</span>
+        ArrowArray ffi_array;
+        ArrowSchema ffi_schema;
+        arrow::ExportRecordBatch(*batch, &amp;ffi_array, &amp;ffi_schema);
+
+        <span class="ty">mosaic</span>::<span class="ty">Writer</span> writer(std::move(cbs), &amp;ffi_schema, {
+            .num_buckets = <span class="num">2</span>,
```
leaves12138: The C++ docs still say to compile with `-std=c++17`, but the examples use C++ designated initializers, which are a C++20 feature. This initializer is also out of declaration order for C++20 (`WriterOptions` declares `compression` before `num_buckets`), so strict compilation fails. Please either switch the docs/build commands to C++20 and order designators by the struct declaration, or avoid designated initializers and use `WriterOptions opts; opts.num_buckets = 2; opts.compression = 1; ...` so the examples match the documented C++17 build command.

docs/index.html
```diff
@@ -0,0 +1,194 @@
...
+                    <tr><td><code>Date</code></td><td>4</td><td>Days since epoch</td></tr>
+                    <tr><td><code>Time</code></td><td>4</td><td>Milliseconds since midnight</td></tr>
+                    <tr><td><code>Char(n)</code></td><td>variable</td><td>Fixed-length string</td></tr>
+                    <tr><td><code>VarChar(n)</code></td><td>variable</td><td>Variable-length string with max length</td></tr>
+                    <tr><td><code>String</code></td><td>variable</td><td>Unbounded UTF-8 string</td></tr>
+                    <tr><td><code>Binary(n)</code></td><td>variable</td><td>Fixed-length byte array</td></tr>
+                    <tr><td><code>VarBinary(n)</code></td><td>variable</td><td>Variable-length byte array with max length</td></tr>
+                    <tr><td><code>Bytes</code></td><td>variable</td><td>Unbounded byte array</td></tr>
+                    <tr><td><code>Decimal(p, s)</code></td><td>8 or variable</td><td>Exact numeric; compact (p&le;18) or large</td></tr>
+                    <tr><td><code>Timestamp(p)</code></td><td>8 or 12</td><td>Millis (p&le;3) or millis + nanos (p&gt;3)</td></tr>
```
leaves12138: This Timestamp summary does not match the implementation: `Timestamp(Millisecond)` and `Timestamp(Microsecond)` are both stored as 8-byte values, while only precision > 6 uses the 12-byte `{millis, nanos_of_milli}` representation. The detailed design page says this correctly; the home page should say something like `millis (p <= 3), micros (p <= 6), or millis + nanos (p > 6)`.

leaves12138 (changes requested):
Thanks for the update. The previous issues look fixed, but I found two remaining doc/code mismatches that should be addressed before merge.

leaves12138 (approved):
Thanks for the update. The previously reported documentation/API mismatches are fixed now. I rechecked the C++ examples against the documented C++17 command, the timestamp description, schema layout, stats examples, Rust Result handling, and Java FloatingPointPrecision usage. Static docs link validation also passed. CI is still queued at review time.

PR #10, opened 2026-05-20
fix: align JNI Arrow C Data ownership

QuakeWang:
## Problem

The Java/JNI writer used inconsistent Arrow C Data ownership semantics.

In the writer open path, Java exported an `ArrowSchema` and released it only after native open returned successfully, while JNI only borrowed the schema. If native open failed, the exported C Data resources could be left unreleased.

The write path had the same issue for exported `ArrowArray` and `ArrowSchema`: JNI moved the structs but only cleared the Java-side structs after a successful import, so import failures could leave exported resources owned by Java without being released. This differed from the C FFI path, which clears the source structs immediately after taking ownership.

Closed writer paths in Java/Python also exported Arrow C Data before native rejected the write.

## Fix

  - Make JNI writer open/write paths consume Arrow C Data structs with `from_raw`, clearing the caller structs immediately.
  - Release Java Arrow C exports in `finally` only when the struct still has an active release callback.
  - Check Java/Python closed writer state before exporting Arrow C Data.
  - Add Java/Python regression coverage for write-after-close.

### DIFF
--- a/java/src/main/java/org/apache/paimon/mosaic/MosaicWriter.java
+++ b/java/src/main/java/org/apache/paimon/mosaic/MosaicWriter.java
@@ -41,19 +41,22 @@ public class MosaicWriter implements AutoCloseable {
     public MosaicWriter(OutputStream outputStream, Schema arrowSchema, WriterOptions options, BufferAllocator allocator) {
         this.allocator = allocator;
         try (ArrowSchema cSchema = ArrowSchema.allocateNew(allocator)) {
-            Data.exportSchema(allocator, arrowSchema, null, cSchema);
-            this.handle = NativeLib.nativeWriterOpen(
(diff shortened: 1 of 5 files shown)


JingsongLi:
@QuakeWang Thanks for your contribution, can you rebase the main?

QuakeWang:
> @QuakeWang Thanks for your contribution, can you rebase the main?

@JingsongLi Done :)

leaves12138 (approved):
Looks good to me. The JNI side now correctly moves ownership of Arrow C Data using from_raw, while Java releases only when the exported structs are still owned on the caller side. The added closed-writer checks also avoid exporting after close. I verified the targeted Java/JNI and Python paths locally.

JingsongLi:
+1

PR #11, opened 2026-05-20
Prepare release yml and tools

JingsongLi:


.github/workflows/release-python.yml
```diff
@@ -0,0 +1,148 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install build tools
+        run: pip install build
+
+      - name: Build sdist
+        working-directory: python
+        run: python -m build --sdist
```
leaves12138: This sdist is built from a fresh checkout before any native library is built, and `setup.py` does not build `mosaic-ffi` from source. In a clean checkout the generated sdist contains neither `libmosaic_ffi.*` nor the Cargo sources needed to build it, so installing from the sdist produces a package that fails at import time when `_ffi.py` cannot find the native library. Please either make the sdist build the Rust library from source (similar to paimon-rust's maturin-based sdist) or do not publish a broken sdist.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,148 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install build tools
+        run: pip install build wheel setuptools
+
+      - name: Build wheel
+        working-directory: python
+        run: python -m build --wheel
```
leaves12138, line 117: Plain `python -m build --wheel` creates native Linux wheels with tags like `cp312-cp312-linux_x86_64`/`linux_aarch64` (locally I reproduced `mosaic_format-0.1.0-cp310-cp310-linux_x86_64.whl`). These are not manylinux wheels and are not suitable for PyPI release. This also unnecessarily restricts a ctypes-only package to the single Python version used by the workflow. Please use a manylinux-aware wheel builder such as `cibuildwheel` + `auditwheel`, or follow paimon-rust's maturin/manylinux approach.

tools/releasing/update_branch_version.sh
```diff
@@ -0,0 +1,60 @@
...
+
+if [ -z "${NEW_VERSION}" ]; then
+	echo "NEW_VERSION is unset"
+	exit 1
+fi
+
+cd ..
+
+#change version in all pom files
+find . -name 'pom.xml' -type f -exec perl -pi -e 's#<version>'$OLD_VERSION'</version>#<version>'$NEW_VERSION'</version>#' {} \;
```
leaves12138, line 56: This version bump only updates `pom.xml`. The PR also adds Rust and Python publishing workflows, whose published versions come from `core/Cargo.toml`, `ffi/Cargo.toml`, `jni/Cargo.toml`, and `python/pyproject.toml`. After creating a new release branch/tag, crates.io/PyPI could still publish the old `0.1.0` version while Maven publishes the new one. Please update all language manifests (and `Cargo.lock` if needed), or add release-time checks that the tag version matches every published package manifest.

leaves12138 (changes requested):
Thanks for preparing the release automation. I compared this with the paimon-rust Python/Rust release flow and paimon-shade's Maven release scripts. The Java/Maven part is close to paimon-shade, but I found blockers in the Python packaging flow and the multi-language version bump before this can reliably publish releases.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,141 @@
...
+
+      - name: Repair wheel (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          pip install auditwheel patchelf
+          auditwheel repair python/dist/*.whl --wheel-dir python/dist/repaired
```
leaves12138: This still builds the Linux wheels on the host runner and only runs `auditwheel repair` afterwards. The repaired tag is therefore bounded by the runner glibc, not by a portable manylinux baseline. Locally, the same flow produced `py3-none-manylinux_2_34_x86_64`, and pip on a glibc 2.32 machine rejected it as unsupported. On `ubuntu-latest` this may be even newer. For PyPI release wheels, please build/repair inside a manylinux container (for example via `cibuildwheel`, or the same manylinux approach used by paimon-rust/maturin) and target a deliberate baseline such as manylinux_2_28 or manylinux2014.

.github/workflows/release-java.yml
```diff
@@ -0,0 +1,145 @@
...
+      - name: Build JNI library
+        run: cargo build --release -p mosaic-jni
+
+      - name: Upload native library
+        uses: actions/upload-artifact@v4
+        with:
+          name: native-${{ matrix.os_name }}-${{ matrix.arch }}
+          path: target/release/${{ matrix.lib_name }}
+
+  deploy-staging:
```
leaves12138, line 84: `deploy-staging` also runs for `workflow_dispatch` on a branch because the only guard is `github.repository == 'apache/paimon-mosaic'`. That can accidentally deploy a branch build (or fail because a SNAPSHOT build needs snapshot credentials while this job configures `apache.releases.https`). Please gate this job on tag refs as well, e.g. `github.repository == 'apache/paimon-mosaic' && startsWith(github.ref, 'refs/tags/')`, or otherwise make the manual workflow require/select a release tag explicitly.

leaves12138 (changes requested):
Thanks for the updates. The Java resource-packaging path is much better now, and I verified locally that the jar can package the native resources and the Java tests can load the JNI library from resources. I still found two release-flow issues that should be fixed before merging.

leaves12138 (approved):
Thanks for the updates. The previous blockers are addressed: the Python Linux wheels now use cibuildwheel with a manylinux_2_28 baseline, and Java staging deploy is gated on tag refs. I also verified the cibuildwheel selector locally and rechecked the Java native-resource packaging/loading path. CI is green. Looks good to me.

PR #12, opened 2026-05-20
Rename Java package to org.apache.paimon.mosaic and update pom.xml

JingsongLi:
- Rename package from io.mosaic to org.apache.paimon.mosaic
- Update groupId to org.apache.paimon.mosaic with Apache parent pom
- Downgrade to Java 1.8 source/target and Arrow 15.0.0
- Add enhanced NativeLib with JAR-embedded native library loading
- Add release profile with GPG signing and enforcer plugin
- Update JNI function names to match new package path

### DIFF
--- a/docs/java-api.html
+++ b/docs/java-api.html
@@ -75,7 +75,7 @@
         &lt;dependency&gt;
             &lt;groupId&gt;org.apache.arrow&lt;/groupId&gt;
             &lt;artifactId&gt;arrow-bom&lt;/artifactId&gt;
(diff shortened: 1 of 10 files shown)


leaves12138 (approved):
Looks good to me. I checked that the Java package rename is consistently reflected in sources, tests, docs, and JNI symbol names, and that no old `io.mosaic` / `Java_io_mosaic` references remain. I also verified Java 8 compatibility locally by building `mosaic-jni` and running `mvn test` with `java.library.path` pointing to the release native library. GitHub CI is green as well.

PR #13, opened 2026-05-20
Add Java snapshot publishing workflow and deploy settings

JingsongLi:


.github/workflows/publish_snapshot.yml
```diff
@@ -0,0 +1,94 @@
...
+
+on:
+  push:
+    branches: [main]
+
+env:
+  JDK_VERSION: 8
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.number || github.run_id }}
```
leaves12138: For `push` events, `github.event.number` is not set, so this falls back to `github.run_id`. Since `run_id` is unique for every run, `cancel-in-progress: true` becomes ineffective. Two quick pushes to `main` can then publish the same `0.1.0-SNAPSHOT` concurrently, and an older commit can finish last and become the latest snapshot metadata. Please group by the branch/ref instead, for example `group: ${{ github.workflow }}-${{ github.ref }}` (or another stable main-branch key).

leaves12138 (changes requested):
Thanks for adding the snapshot publishing workflow. The Java/JNI/package changes look good and validation passes, but the snapshot workflow concurrency group should be fixed before merge to avoid stale snapshot publications.

leaves12138 (approved):
I pushed a follow-up commit to make the snapshot workflow concurrency key stable for the main branch (`${{ github.workflow }}-${{ github.ref }}`). This addresses my previous concern about stale concurrent snapshot deployments. Rechecked the latest branch contents; only that workflow line changed. CI is still running on the new commit at review time.

PR #14, opened 2026-05-20
fix: free native reader handle on constructor failure

QuakeWang:
`MosaicReader.open()` acquired the native reader handle before the Java wrapper finished constructing. If schema export/import failed in the constructor, object creation aborted and `close()` never ran, leaking the native handle.

This change frees the native handle immediately when construction fails, then rethrows the original exception. A regression test covers the failure path.


JingsongLi:
+1

PR #15, opened 2026-05-20
Remove release-cpp.yml

JingsongLi:
CPP is usually packaged based on the source code, so we don't need to release binary packages.


leaves12138 (approved):
Looks good to me. This PR only removes the standalone C++ release workflow. I checked that no release-cpp/GitHub Release packaging references remain, the C++ CI test workflow is still present, and all current checks are green.

PR #18, opened 2026-05-20
fix: honor projection for monolithic buckets

QuakeWang:
Fix monolithic bucket reads to respect the projected column set when building `RecordBatch`, so partial projection no longer returns extra columns. Add a regression test for the `compression=none`, `num_buckets=2`, single-column projection case.


JingsongLi:
+1

PR #19, opened 2026-05-20
fix: preserve row count for empty projection

QuakeWang:
Fix empty projection reads to build a valid zero-column `RecordBatch` with the row group row count, so `columns=[]` no longer fails with Arrow's missing row count error. Add a regression test for the empty projected columns case.


JingsongLi:
+1

PR #20, opened 2026-05-20
Reject unsupported types in stats_columns at writer creation time

JingsongLi:
Previously unsupported types were silently filtered, making the returned statistics list not correspond positionally to the input stats_columns. Now the writer returns an error if any stats_columns index is out of range or refers to a type that does not support statistics. Also adds Javadoc/docstring comments clarifying that getRowGroupStatistics returns entries in the same order as stats_columns.

### DIFF
--- a/core/src/writer.rs
+++ b/core/src/writer.rs
@@ -99,10 +99,10 @@ impl<S: OutputFile> MosaicWriter<S> {
     pub fn new(out: S, schema: &Schema, options: WriterOptions) -> io::Result<Self> {
         let mosaic_schema = MosaicSchema::from_arrow(schema, options.num_buckets)
             .map_err(|e| io::Error::new(io::ErrorKind::InvalidInput, e))?;
-        Ok(Self::from_mosaic_schema(out, mosaic_schema, options))
+        Self::from_mosaic_schema(out, mosaic_schema, options)
     }
 
-    pub fn from_mosaic_schema(out: S, schema: MosaicSchema, options: WriterOptions) -> Self {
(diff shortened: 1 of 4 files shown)


leaves12138 (changes requested):
Thanks for the fix. The functional change looks good to me: rejecting out-of-range and unsupported `stats_columns` at writer creation time is the right behavior, and the FFI/JNI/Python paths propagate the open failure as expected.

I found one CI-blocking issue: `cargo fmt --check` fails on `core/src/writer.rs` for the new `from_mosaic_schema` signature and the two long `format!` calls. Please run `cargo fmt` and push the formatted result.

Local validation I ran:
- `cargo test -q -p mosaic-core --lib stats_columns`
- `cargo build -q -p mosaic-ffi --release`
- `cargo build -q -p mosaic-jni --release`
- Python stats tests plus an unsupported/out-of-range `stats_columns` smoke test
- Java `MosaicRoundtripTest#testWriterStats*` and `MosaicRoundtripTest#testReaderStats*`
- `cargo test -q` passed for the Rust workspace
- `cargo fmt --check` failed as noted above

leaves12138 (approved):
LGTM. The formatting issue from my previous review is fixed.

I rechecked the latest head and ran:
- `cargo fmt --check`
- `cargo test -q -p mosaic-core --lib stats_columns`
- `cargo build -q -p mosaic-ffi --release`
- `cargo build -q -p mosaic-jni --release`

All passed locally. GitHub checks were still queued when I reviewed the updated commit.

PR #21, opened 2026-05-20
feat: expose timestamp nanos as Arrow timestamp(ns)

QuakeWang:
## Problem

Mosaic exposes TIMESTAMP precision > 6 as a legacy Arrow Struct (`millis`, `nanos_of_milli`) instead of standard Arrow timestamp(ns). This leaks the internal 12-byte storage representation into the Arrow API and loses timezone metadata for TIMESTAMP_LTZ precision > 6.

## Fix

Keep the Mosaic file format unchanged, but expose high-precision timestamps through standard Arrow timestamp(ns):

- Deserialize precision > 6 timestamps as `Timestamp(Nanosecond, timezone)`.
- Read and write `TimestampNanosecondArray` using the existing `millis + nanos_of_milli` storage encoding.
- Preserve TIMESTAMP_LTZ timezone metadata.
- Keep legacy Struct input compatibility with validation.
- Add Rust, Python, C++, and Java timestamp(ns) roundtrip coverage.

Precision 7/8 is normalized to Arrow timestamp(ns), since Arrow does not preserve decimal timestamp precision separately.


JingsongLi:
+1

PR #22, opened 2026-05-20
Remove logical index, add name-based project() and batch stats API

JingsongLi:
- Remove globalColumnIndex from schema serialization; columns are now always stored and returned in name-sorted order
- Add reader-level project(column_names) API that controls output column order (supports empty projection for count-only reads)
- Replace per-element stat FFI/JNI calls with batch array methods (4 arrays: names, null_counts, mins, maxs) to reduce crossing overhead
- Stats API returns Map<String, ColumnStatistics> keyed by column name (Java/Python) or vector<ColumnStatistics> with column_name field (C++)
- WriterOptions.stats_columns now takes column names instead of indices
- Add projection order and empty projection tests across all languages
- Update all API docs and C++ tests to reflect the new interfaces

### DIFF
--- a/include/mosaic.hpp
+++ b/include/mosaic.hpp
@@ -89,7 +89,7 @@ struct WriterOptions {
     uint64_t row_group_max_size = 256ULL * 1024 * 1024;
     uint32_t max_dict_total_bytes = 32 * 1024;
     uint32_t max_dict_entries = 255;
-    const uint32_t* stats_columns = nullptr;
+    const char* const* stats_columns = nullptr;
     uint32_t num_stats_columns = 0;
     uint32_t page_size_threshold = 32 * 1024;
 };
@@ -97,13 +97,16 @@ struct WriterOptions {
 // ======================== Statistics ========================
 
 struct ColumnStatistics {
-    uint32_t column_index;
+    std::string column_name;
     uint64_t null_count;
     std::vector<uint8_t> min_value;
     std::vector<uint8_t> max_value;
     bool has_min_max() const { return !min_value.empty(); }
 };
 
+/// Columns are stored in name-sorted order regardless of input schema order.
+/// The reader returns columns in storage (name-sorted) order by default;
+/// use Reader::set_projection() to control output column order.
 class Writer {
 public:
     /// Construct a writer. `arrow_schema` is a pointer to an ArrowSchema (Arrow C Data Interface).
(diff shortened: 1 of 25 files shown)


python/mosaic/mosaic.py
```diff
@@ -59,6 +59,32 @@ def _check_error(msg="operation failed"):
...
+    min_lens = (ctypes.c_size_t * n)()
+    max_ptrs = (ctypes.POINTER(ctypes.c_uint8) * n)()
+    max_lens = (ctypes.c_size_t * n)()
+    rc = stats_fn(handle, rg_index, names, null_counts, min_ptrs, min_lens, max_ptrs, max_lens)
+    if rc != 0:
+        _check_error("row_group_stats failed")
+    result = {}
+    for i in range(n):
+        col_name = names[i].decode("utf-8")
+        min_val = ctypes.string_at(min_ptrs[i], min_lens[i]) if min_ptrs[i] and min_lens[i] > 0 else None
```
leaves12138, line 82: This treats a valid zero-length stats value as missing. For a UTF8 stats column whose minimum is the empty string, the native side returns a non-null pointer with `min_lens[i] == 0`, but this branch converts it to `None`, so `has_min_max` becomes false even though the row group is not all-null. Please use pointer presence to distinguish missing values from empty values, e.g. call `ctypes.string_at(min_ptrs[i], min_lens[i])` whenever `min_ptrs[i]` is non-null, even if the length is zero. The same applies to `max_val`, and it would be good to add a Python test with stats over `["", "b"]`.

include/mosaic.hpp
```diff
@@ -295,22 +298,30 @@ class Reader {
...
-            size_t min_len = 0, max_len = 0;
-            const uint8_t* min_ptr = mosaic_reader_row_group_stat_min(handle_, rg_index, i, &min_len);
-            const uint8_t* max_ptr = mosaic_reader_row_group_stat_max(handle_, rg_index, i, &max_len);
-            if (min_ptr && min_len > 0)
-                s.min_value.assign(min_ptr, min_ptr + min_len);
-            if (max_ptr && max_len > 0)
-                s.max_value.assign(max_ptr, max_ptr + max_len);
+            s.column_name = names[i] ? names[i] : "";
+            s.null_count = null_counts[i];
+            if (min_ptrs[i] && min_lens[i] > 0)
```
leaves12138, line 321: Same zero-length stats issue exists in the C++ wrapper. Empty string min/max values have length 0, so this skips assigning them; then `ColumnStatistics::has_min_max()` (based on `!min_value.empty()`) reports false. The wrapper needs to preserve min/max presence separately from byte length, for example by adding explicit `has_min`/`has_max` flags (or optional byte vectors) and assigning even when `*_lens[i] == 0` if the pointer is non-null. Please fix both writer and reader stats paths and add a C++ test for an empty-string min/max.

leaves12138 (changes requested):
Thanks for the API cleanup. The name-based projection and batch stats API generally look good, and I verified the Rust workspace locally with `cargo fmt --all -- --check` and `cargo test --workspace`. I found one cross-language stats correctness issue that should be fixed before merging: zero-length string min/max values are currently indistinguishable from missing min/max in the Python and C++ bindings.

PR #23, opened 2026-05-20
Preserve original column order in reader and fix stats for empty strings

JingsongLi:
- Store original column order as delta+zigzag encoded permutation in schema block
- Reader returns columns in original input order by default (not name-sorted)
- Add project() API to select and reorder output columns
- Fix empty string min/max statistics handling across all language bindings
- Add stats_columns writer option and per-row-group statistics API
- Update docs and add tests for all languages
- Fix clippy warnings (type_complexity, needless_borrows, iter_cloned_collect)


leaves12138 (changes requested):
Thanks for the update. The default read/schema export behavior now preserves the writer's column order, projection still keeps the requested order, and the empty-string min/max handling is covered across bindings.

I ran locally:
- `cargo test --workspace`
- `cargo build --release -p mosaic-ffi`
- Python tests with `MOSAIC_LIB_PATH=.../target/release`: 28 passed
- `cargo build -p mosaic-jni --release`
- Java `mvn test -DargLine=-Djava.library.path=.../target/release`: 25 passed

I could not run the local C++ test because this environment does not have the Arrow C++ CMake package (`ArrowConfig.cmake` / `arrow-config.cmake`); the GitHub C++ CI is still queued at the time of review.

One remaining robustness issue before merge: please validate `original_order` during schema deserialization. After decoding the appended order, it should be rejected unless it is exactly a permutation of `0..num_columns`. Right now a corrupted schema block can deserialize successfully with duplicate or out-of-range indexes. That breaks the `MosaicSchema` invariant and can surface later as a failed default read, a silently incomplete/incorrect default output order for duplicates, or a panic in schema export paths that index `schema.columns[i]` and rely on `catch_unwind`.

Suggested fix: immediately after decoding `original_order`, check `len == num_columns`, every entry is `< num_columns`, and every entry appears exactly once; return `io::ErrorKind::InvalidData` otherwise. It would also be good to add robustness tests for duplicate and out-of-range original-order metadata.

leaves12138 (approved):
Thanks for the update. The `original_order` metadata is now validated as a proper permutation during schema deserialization, and the duplicate/out-of-range tests cover the corruption cases I was worried about.

I re-reviewed the latest head `ea7ea27c75c39acf3f5e507921b9f4c746d2c29b` and ran:
- `cargo test --workspace`

All passed locally. The GitHub CI jobs were still queued when I checked, so please wait for them before merge.

PR #24, opened 2026-05-21
fix: schema layout version compatibility

QuakeWang:
## Problem

Schema block layout changed while the file footer version stayed at v1. Existing v1 files can be parsed with the new layout, causing the legacy logical index to be read as a type header and potentially corrupting schema metadata silently.

## Fix

Bump the current file version to v2 and keep v1 as the legacy format. Route schema deserialization by footer version, parse v1 with the legacy logical-index layout, and add a v2 schema layout marker for the current format.

Also add regression coverage for legacy v1 schema decoding.

### DIFF
--- a/core/src/reader.rs
+++ b/core/src/reader.rs
@@ -264,7 +264,7 @@ impl<I: InputFile> MosaicReader<I> {
         }
 
         let version = footer[25];
-        if version != VERSION {
+        if version != LEGACY_VERSION && version != VERSION {
             return Err(io::Error::new(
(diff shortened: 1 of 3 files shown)


JingsongLi:
We don't need to do this. This format has not been publicly released yet, so there's no need to consider compatibility issues.

QuakeWang:
> We don't need to do this. This format has not been publicly released yet, so there's no need to consider compatibility issues.

Got it. I will close this PR.

PR #25, opened 2026-05-21
fix: preserve empty projection in read_table

QuakeWang:
## Problem

`read_table(..., columns=[])` currently skips projection because the Python wrapper checks `if columns:`. This makes an explicit empty projection behave the same as `columns=None`, so the convenience API returns all columns instead of a zero-column table with the row count preserved.

## Fix

Change the wrapper to call `reader.project(columns)` whenever `columns is not None`, preserving the existing three-state projection semantics:

- `None`: read all columns
- non-empty list: read selected columns
- empty list: read zero columns and preserve row count

Add a regression test for `read_table(..., columns=[])`.


JingsongLi (approved):
+1

PR #28, opened 2026-05-21
fix: preserve Python projection schema for empty reads

QuakeWang:
## Problem

Python `read_table(..., columns=...)` loses the requested projection when reading a Mosaic file with zero row groups.

`MosaicReader.read_all()` normally uses the first returned batch schema, so projected reads with row groups return the expected schema. For zero row group files there is no batch, and the fallback used the full reader schema. As a result, `columns=[]` and projected column reads returned an empty table with the full file schema.

## Fix

Keep the core/FFI behavior unchanged and make the Python fallback projection-aware:

  * Cache the projected schema in `MosaicReader.project()` after native projection succeeds.
  * Use the cached projected schema in `read_all()` when there are no batches.
  * Preserve `reader.schema` as the full file schema.
  * Match native duplicate projection behavior by keeping the first occurrence only.
  * Add Python coverage for zero row group reads with empty, single-column, and duplicate projections.

### DIFF
--- a/python/mosaic/mosaic.py
+++ b/python/mosaic/mosaic.py
@@ -284,6 +284,7 @@ class MosaicReader:
         if rc != 0:
             _check_error("export_schema failed")
         self._schema = pa.Schema._import_from_c(schema_ptr)
+        self._projected_schema = None
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


JingsongLi:
+1

PR #32, opened 2026-05-22
docs: align release process with RC tags

QuakeWang:
## Summary

- use a stable release branch (`release-0.1`) instead of RC-specific branches
- move dependency license manifest generation before cutting the release branch
- use explicit RC/final tag variables and verify signed tags
- generate source artifacts from the same commit as the RC tag

## Context

This aligns the paimon-mosaic release guide with the paimon-rust release workflow, where RC attempts are represented by tags such as `v0.1.0-rc1` rather than branch names.


XiaoHongbo-Hope:
+1

PR #33, opened 2026-05-22
Add support for LZ4 codec

ArnavBalyan:
- Currently only zstd is supported, however LZ4 may be preferred for GPU workloads for fast decode/encode
 - Add support for the same.

### DIFF
--- a/core/src/spec.rs
+++ b/core/src/spec.rs
@@ -21,6 +21,7 @@ pub const FOOTER_SIZE: usize = 32;
 
 pub const COMPRESSION_NONE: u8 = 0;
 pub const COMPRESSION_ZSTD: u8 = 1;
+pub const COMPRESSION_LZ4: u8 = 2;
 
 pub const ENCODING_PLAIN: u8 = 0;
 pub const ENCODING_CONST: u8 = 1;
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -27,6 +27,7 @@ crate-type = ["rlib"]
 
 [dependencies]
 zstd = "0.13"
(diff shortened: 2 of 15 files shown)


ArnavBalyan:
cc @JingsongLi thanks!

core/src/spec.rs
```diff
@@ -21,6 +21,7 @@ pub const FOOTER_SIZE: usize = 32;
 
 pub const COMPRESSION_NONE: u8 = 0;
 pub const COMPRESSION_ZSTD: u8 = 1;
+pub const COMPRESSION_LZ4: u8 = 2;
```
QuakeWang, line 24: Adding a new on-disk compression id makes this a format/API-level change, but the public bindings and docs are not updated consistently. Java exposes COMPRESSION_LZ4, while Python still only exposes NONE/ZSTD, C++ has no named constant, and the format/API docs still say only `0 = none, 1 = zstd` and that paged buckets require ZSTD.

Please update the Python/C++ public APIs, docs, and at least add cross-language roundtrip coverage for `compression = 2`, so the new codec has one consistent meaning across all supported entry points.

ArnavBalyan: Did not know we are at a public stage yet, kindly look again

QuakeWang (approved):
LGTM

cc @JingsongLi

JingsongLi (changes requested):
I haven't seen any actual more effective aspects of LZ4 yet. Have you really tested it in practice?

ArnavBalyan:
Will revisit this with benchmark/numbers closing now

PR #35, opened 2026-05-22
Should we enable issues and discussions? 

ArnavBalyan:
- Looks like issues is disabled but readme refers to it.
 - Maybe we can remove it from the readme or enable it on the repo settings?


ArnavBalyan:
cc @JingsongLi @XiaoHongbo-Hope would it be possible to open issues or maybe this section could be cleaned up thanks!

JingsongLi:
We should open issue for this repo.

PR #38, opened 2026-05-23
ci: fix release signing and macOS wheel repair

QuakeWang:
## Summary

- Added loopback pinentry arguments for Java release artifact signing in CI.
- Updated the macOS Python wheel repair step to require the `arm64` architecture explicitly.
- Kept the release workflow scope limited to the two failures observed in RC1.


XiaoHongbo-Hope:
+1

PR #39, opened 2026-05-23
ci: fix macOS Python wheel platform tag

QuakeWang:
Fix the macOS Python release wheel build to produce an arm64-specific wheel tag.


XiaoHongbo-Hope:
+1

PR #42, opened 2026-05-24
ci: publish Python release after Rust and Java

QuakeWang:
## Summary

- Add a thin release orchestrator workflow for tag-based releases
- Keep Rust, Java, and Python wheel release workflows reusable and manually runnable
- Move Python PyPI/TestPyPI publishing into a final job that runs only after Rust, Java, and Python wheel jobs succeed
- Update release docs to describe the new ordering

### DIFF
--- a/.github/workflows/release-java.yml
+++ b/.github/workflows/release-java.yml
@@ -19,18 +19,15 @@
 name: Release Java
 
 on:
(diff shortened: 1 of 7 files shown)


XiaoHongbo-Hope:
+1

PR #43, opened 2026-05-24
ci: fix Java Linux native glibc baseline

QuakeWang:
### What changed

This updates Java native release and snapshot workflows to build Linux GNU JNI artifacts with `cargo-zigbuild` using a fixed glibc baseline:

- `x86_64-unknown-linux-gnu.2.17`
- `aarch64-unknown-linux-gnu.2.17`

The workflows now also verify the generated Linux native library with `readelf --version-info` and fail if the required `GLIBC_*` version is higher than `GLIBC_2.17`.

Non-Linux artifacts continue to use regular `cargo build --target`.

### Why

The previous workflows built Linux JNI libraries directly on the GitHub runner. On newer Ubuntu runners, this can produce `.so` artifacts requiring newer glibc symbols, making them fail to load on common enterprise Linux distributions with older glibc versions.

### DIFF
--- a/.github/workflows/publish_snapshot.yml
+++ b/.github/workflows/publish_snapshot.yml
@@ -77,14 +77,60 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-${{ matrix.target }}-cargo-
 
+      - name: Setup Python
+        if: matrix.os_name == 'linux'
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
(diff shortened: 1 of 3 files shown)


XiaoHongbo-Hope:
nit: would be nice to document the glibc 2.17 baseline somewhere user-facing (e.g. release notes /
  README) so users know the compatibility range. Not blocking

XiaoHongbo-Hope:
+1

PR #45, opened 2026-05-25
Fix BPE non-determinism and add comprehensive test suite

JingsongLi:
Fix a non-determinism bug in BPE vocabulary builder where HashMap iteration order caused different byte output for identical input data when column names share common substrings. The fix uses the pair key as a deterministic tie-breaker when multiple pairs have equal frequency.

Add 150+ new tests across Rust, Java, and Python covering:
- Massive data volumes (up to 10M rows, 1000 columns)
- 20 data pattern varieties (sawtooth, fibonacci, unicode, sparse, etc.)
- Encoding/compression/projection interactions (30 tests)
- Determinism and re-roundtrip stability (18 tests)
- Concurrent multi-threaded reading (5 tests)
- Binary format byte-level verification (10 tests)
- Statistics min/max/null_count accuracy (11 tests)
- Cross-language Rust/Java/Python interoperability (25 tests)


leaves12138 (approved):
Thanks for the fix. The BPE vocabulary tie-breaker now uses the pair key together with frequency, so equal-frequency pairs no longer depend on `HashMap` iteration order.

I reviewed the CI changes and the added Rust/Java/Python coverage, including the shared-substring determinism regression and cross-language interop flow. I also ran the targeted Rust checks locally:

```text
cargo test -p paimon-mosaic-core bpe -- --nocapture
cargo test -p paimon-mosaic-core --test determinism_test test_deterministic_shared_substring_columns -- --nocapture
```

Both passed, and GitHub CI is green. +1.

PR #46, opened 2026-05-25
ci: print line coverage summary in CI logs

jianguotian:
## Summary

- Replace `cargo test` with `cargo llvm-cov --summary-only` in the rust-test job so each CI run prints a per-file line/region/function coverage table directly in the Actions log.
- Install `cargo-llvm-cov` via `taiki-e/install-action` before running coverage.
- No external service (Codecov etc.) dependency — coverage is visible by reading the workflow output.


JingsongLi:
+1

PR #48, opened 2026-05-28
release: fix source tarball and Java verification paths

QuakeWang:
The source release script used rsync and the platform tar command, which can include macOS filesystem metadata such as AppleDouble files, pax headers, and com.apple.* extended attributes in the source tarball.

This PR generates the source archive from the committed Git tree with git archive and gzip -n, avoiding filesystem metadata in the release tarball.

The release verification guide also copied the JNI library into java/src/main/resources/native/, but NativeLib loads libraries from /native/{os}/{arch}/. This PR updates the guide to copy each platform library into the directory layout used by NativeLib and the Java release workflow.

### DIFF
--- a/docs/creating-a-release.html
+++ b/docs/creating-a-release.html
@@ -178,7 +178,7 @@ python3 tools/dependencies.py generate
 git add DEPENDENCIES.rust.tsv core/DEPENDENCIES.rust.tsv ffi/DEPENDENCIES.rust.tsv jni/DEPENDENCIES.rust.tsv
 git commit -m "chore: update dependency list for release ${RELEASE_VERSION}"
 git push origin main</code></pre>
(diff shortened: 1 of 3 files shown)


XiaoHongbo-Hope:
One small doc clarity suggestion: the Java section first runs `cargo build
  --release -p paimon-mosaic-jni`, then lists copy commands for all platforms from `target/release/`.
  Could we explicitly say verifiers should run only the snippet for the current platform?

XiaoHongbo-Hope:
Nit: `docs/creating-a-release.html` still says `create_source_release.sh` builds from a git clone. Since
  this PR switches to `git archive`, maybe update it to "from the committed Git tree".

QuakeWang:
@XiaoHongbo-Hope Thanks, I have fixed. The Java verification section now explicitly says to run only the snippet for the current platform. I also updated the release guide wording to say the source archive is built from the committed Git tree.

XiaoHongbo-Hope:
+1

PR #49, opened 2026-06-02
release: fix Java NOTICE year

QuakeWang:
The Java release artifact generated META-INF/NOTICE from maven-remote-resources-plugin. The project inherited Apache parent 23's default output timestamp from 2020 while java/pom.xml declared inceptionYear as 2025, which produced an invalid copyright range: 2025-2020.

This PR aligns the project NOTICE year with the first Apache Paimon Mosaic release year and sets the Maven output timestamp to 2026 so generated Java artifacts contain a valid NOTICE file.


XiaoHongbo-Hope:
+1

PR #50, opened 2026-06-03
release: decouple Python publish from Rust and Java release jobs

JingsongLi:
Python wheels have no runtime dependency on Rust or Java artifacts, so the publish step only needs to wait for the wheel build to finish.


XiaoHongbo-Hope:
+1

PR #52, opened 2026-06-05
Support ARRAY type with ORC-style flattened columnar storage

JingsongLi:
## Summary

Add ARRAY (List) type support using **ORC-style flattened columnar storage**. Each ARRAY column is decomposed into physical columns within the same bucket — a lengths column (INT32) and a values column (element type) — both first-class columns sharing the same encoding infrastructure (DICT/CONST/PLAIN/ALL_NULL).

Key design points:
- **No sub-bucket or nested container** — lengths and values are peer columns in the bucket, like ORC's LENGTH stream + child column
- **Nested arrays** (`ARRAY<ARRAY<INT>>`) expand recursively into N+C physical columns, all within the same bucket
- **Full column encoding** — element values benefit from dictionary compression across all arrays in the column
- **Parquet's rep/def levels were considered but not adopted** due to implementation complexity and limited benefit for common 1-2 level nesting (see design doc for rationale)

## Storage Format

```
Monolithic bucket with ARRAY<INT>:
  varint(num_primary=N) + varint(num_children=C)
  + varint(child_element_count) per child
  + [encoding flags, null bitmaps, column data for all N+C columns]

Paged bucket:
  u16(num_children) + u32(element_count) per child
  + directory with N+C entries
  + independent compressed slots for all N+C columns

ARRAY<ARRAY<INT>> expands to 3 physical columns:
  [outer_lengths (INT32)] [inner_lengths (INT32)] [leaf_values (INT32)]
```

## Changes

- **core/types.rs**: Type byte 18 for ARRAY, recursive schema serialize/deserialize with child field name preservation
- **core/bucket_writer.rs**: `expand_col_types` expands ARRAY into physical columns; `extract_list_lengths` + `flatten_list_values` canonicalize null rows; `take_array` for filtering child values
- **core/bucket_reader.rs**: Per-column row counts (`col_num_rows`); `reassemble_list_columns` rebuilds ListArrays from physical columns
- **core/reader.rs**: Paged path handles N+C physical columns for directory, slot sizing, projection, and round-2 reads; bounds checks for corrupted data
- **docs/design.html**: ARRAY storage format spec, Mosaic vs ORC comparison table, rationale for not using Parquet rep/def levels
- **docs/java-api.html, python-api.html, cpp-api.html**: ARRAY usage examples

## Test Plan

- [x] Rust: 10 array tests (basic, null elements, strings, nested, all-null, mixed columns, large batch, paged layout, null row offset preservation, paged partial projection)
- [x] Rust: 229 total tests pass (125 lib + 10 array + 94 integration)
- [x] Java: 27 tests pass (including `testArrayType`)
- [x] Python: 35 tests pass (including 4 array tests)
- [x] C++: 3 array tests added
- [x] Clippy: 0 warnings (`-D warnings`)
- [x] Robustness: `test_corrupted_compressed_data` passes (bounds checks)


leaves12138 (changes requested):
Thanks for adding ARRAY support. I found one correctness blocker around Arrow list null slots.

The writer currently stores the list length as `offsets[i + 1] - offsets[i]` for every row, including null list rows, but it flattens all child values from `offsets[0]..offsets[len]`. During read, `reassemble_list_array` treats a null list length as `0` when rebuilding offsets.

This loses the child-value range of null list slots. Arrow allows a null list slot to have a non-empty offset range; consumers must ignore that slot logically, but preserving the offsets is necessary so following rows keep pointing to the right child values.

I reproduced it with a temporary Rust test using offsets `[0, 2, 4, 5]`, validity `[valid, null, valid]`, and values `[1, 2, 99, 100, 5]`. After round-trip, the result offsets became `[0, 2, 2, 3]`, so row 2 reads from the wrong child position instead of preserving `[5]`.

A minimal shape of the reproducer is:

```rust
let offsets = OffsetBuffer::new(ScalarBuffer::from(vec![0i32, 2, 4, 5]));
let values = Arc::new(Int32Array::from(vec![1, 2, 99, 100, 5])) as ArrayRef;
let nulls = Some(NullBuffer::new(BooleanBuffer::new(Buffer::from(vec![0b0000_0101]), 0, 3)));
let array = ListArray::new(element_field.clone(), offsets, values, nulls);
```

Expected row 2 after round-trip is `[5]`, but with the current encoding row 2 is shifted.

Please preserve the original offset deltas for null list slots as well when rebuilding offsets (or otherwise keep the flattened values and offsets consistent), and add a test for a null list row with a non-empty offset range.

Validation I ran:

- `cargo fmt --all -- --check` passes.
- `cargo test -p paimon-mosaic-core --test array_type_test` passes existing 7 tests.
- `cargo test -p paimon-mosaic-core` is blocked by existing interop tests requiring `/tmp/mosaic_interop/*` files, unrelated to this change.

core/src/bucket_writer.rs
```diff
@@ -566,6 +643,56 @@ impl BucketWriter {
...
         }
 
+        for lc in &self.list_columns {
+            let col = lc.physical_col;
+            let values_paged = lc.values_writer.finish_paged();
+
+            let mut composite = Vec::new();
+            varint::encode(&mut composite, lc.total_elements as u32);
+
+            if let Some(ref lengths_page) = column_pages[col] {
```
QuakeWang: This writes only the raw lengths column page data into the ARRAY composite slot, but `parse_list_column_slot()` later passes it to `parse_simple_column_slot()`, which expects a full page_content buffer starting with `[encoding, flags, const/meta...]`. With paged layout forced (`page_size_threshold = 1`), ARRAY roundtrip fails with `InvalidData: column page truncated: plain fixed-width data`. Please either store a complete simple-column page_content for lengths/values, or teach the reader to parse the exact raw page format written here, and add a forced-paged ARRAY regression test.

core/src/types.rs
```diff
@@ -258,6 +263,10 @@ pub fn deserialize_field(name: &str, buf: &[u8], pos: &mut usize) -> Result<Fiel
                 DataType::Timestamp(TimeUnit::Nanosecond, Some(tz))
             }
         }
+        18 => {
+            let element_field = deserialize_field("element", buf, pos)?;
```
QuakeWang: This hard-codes the deserialized list child field name to `"element"`, so schema roundtrip does not preserve Arrow's original child field name, e.g. `List(Field { name: "custom_item", ... })` becomes `List(Field { name: "element", ... })`. If Mosaic's schema contract is intended to preserve Arrow schema fidelity, the serialized ARRAY descriptor needs to include the child field name, or the limitation should be documented and tested explicitly.

leaves12138 (changes requested):
Thanks for the update. I re-checked the latest head (`7700b0e8d35863312365ff24387c001f16f91366`) carefully.

The existing array tests now pass (`cargo test -p paimon-mosaic-core --test array_type_test`, 8 tests), and `cargo fmt --all -- --check` passes. However, the previous correctness blocker is still reproducible.

The code still writes list lengths as `offsets[i + 1] - offsets[i]` for every row, including null list rows, and still flattens the whole child range from `offsets[0]..offsets[len]`. But `reassemble_list_array` still rebuilds offsets with length `0` for null rows:

```rust
let len = if lengths_arr.is_null(i) {
    0
} else {
    lengths_arr.value(i)
};
```

That keeps the logical null marker, but it drops the child range of null list slots and shifts subsequent rows. The reproducer still fails on the latest head:

```rust
let offsets = OffsetBuffer::new(ScalarBuffer::from(vec![0i32, 2, 4, 5]));
let values = Arc::new(Int32Array::from(vec![1, 2, 99, 100, 5])) as ArrayRef;
let nulls = Some(NullBuffer::new(BooleanBuffer::new(Buffer::from(vec![0b0000_0101]), 0, 3)));
let array = ListArray::new(element_field.clone(), offsets, values, nulls);
```

Expected offsets after round-trip: `[0, 2, 4, 5]` and row 2 should read `[5]`. Actual offsets are `[0, 2, 2, 3]`, so row 2 is shifted.

Please either preserve the original offset deltas for null list rows when rebuilding offsets, or filter out the child values for null rows at write time so that flattened values and rebuilt offsets remain consistent. Please also add this exact kind of test; the current builder-based null-list tests do not cover null slots with non-empty offset ranges.

I also noticed the new list sub-bucket/page parsing uses unchecked slices such as `self.data[pos..pos + sub_blob_size]` and `page_content[pos..pos + lengths_page_size]`. Existing readers generally return `io::Error` for malformed/truncated input; these new paths can panic on corrupt/truncated files. Please add bounds checks before slicing.

leaves12138 (changes requested):
Thanks for the update. The flattened physical-column layout is much cleaner, but I found two correctness issues that still need to be fixed before this can merge.

1. Null list rows still corrupt offsets and subsequent values.

`extract_list_lengths` writes the real Arrow offset delta for every row, including null list rows, and `flatten_list_values` writes the full child range. However `reassemble_list_array` still rebuilds offsets with length `0` whenever the length slot is null. For a valid/null/valid list array whose null row owns child values, the following focused regression fails on this PR head (`2bd4a72`):

```rust
#[test]
fn test_array_null_row_preserves_child_offsets_regression() {
    use arrow_buffer::{BooleanBuffer, Buffer, NullBuffer, OffsetBuffer, ScalarBuffer};

    let element_field = Arc::new(Field::new("item", DataType::Int32, true));
        "arr",
        DataType::List(element_field.clone()),
        true,
    )]);

    let offsets = OffsetBuffer::new(ScalarBuffer::from(vec![0i32, 2, 4, 5]));
    let values = Arc::new(Int32Array::from(vec![1, 2, 99, 100, 5])) as ArrayRef;
    let nulls = Some(NullBuffer::new(BooleanBuffer::new(
        Buffer::from(vec![0b0000_0101]),
        0,
        3,
    )));
    let array = ListArray::new(element_field, offsets, values, nulls);
    let batch = RecordBatch::try_new(Arc::new(schema.clone()), vec![Arc::new(array.clone())]).unwrap();

    let result = roundtrip(&schema, &[batch]);
    let result_col = result[0].column(0).as_any().downcast_ref::<ListArray>().unwrap();

    assert_eq!(result_col.value_offsets(), &[0, 2, 4, 5]);
    assert_eq!(&result_col.value(0), &array.value(0));
    assert!(result_col.is_null(1));
    assert_eq!(&result_col.value(2), &array.value(2));
}
```

Actual result is `value_offsets() == [0, 2, 2, 3]`, so row 2 reads the first hidden value from the null row instead of `5`. Please either preserve the stored lengths even for null list rows when rebuilding offsets, or canonicalize/filter null-row child ranges on the write path consistently.

2. Paged partial projection does not account for ARRAY physical child columns.

In the paged partial-projection path, `num_columns` is the number of physical columns after ARRAY expansion, but the code indexes `global_indices[i]` and allocates `slot_locations` using the number of logical columns. This panics as soon as a projected ARRAY column requires its child physical slot:

```rust
#[test]
fn test_project_array_from_paged_bucket_regression() {
    let element_field = Arc::new(Field::new("item", DataType::Int32, true));
    let schema = Schema::new(vec![
        Field::new("id", DataType::Int64, false),
        Field::new("arr", DataType::List(element_field.clone()), true),
    ]);

    let ids = Int64Array::from(vec![1, 2, 3]);
    let mut list_builder = ListBuilder::new(Int32Builder::new());
    list_builder.values().append_value(10);
    list_builder.values().append_value(20);
    list_builder.append(true);
    list_builder.values().append_value(30);
    list_builder.append(true);
    list_builder.values().append_value(40);
    list_builder.append(true);
    let arr = list_builder.finish();
    let batch = RecordBatch::try_new(Arc::new(schema.clone()), vec![Arc::new(ids), Arc::new(arr)]).unwrap();

    let options = WriterOptions {
        num_buckets: 1,
        page_size_threshold: 1,
        ..WriterOptions::default()
    };
    let mut writer = MosaicWriter::new(out, &schema, options).unwrap();
    writer.write_batch(&batch).unwrap();
    writer.close().unwrap();

    let data = writer.output().buf.clone();
    let input = ByteArrayInputFile { data: data.clone() };
    let reader = MosaicReader::new(input, data.len() as u64).unwrap();
    let sorted_arr_idx = reader.schema().columns.iter().position(|c| c.name == "arr").unwrap();
    let mut rg_reader = reader.row_group_reader_projected(0, &[sorted_arr_idx]).unwrap();
    let projected = rg_reader.read_columns().unwrap();
    let result_arr = projected.column(0).as_any().downcast_ref::<ListArray>().unwrap();
    assert_eq!(result_arr.value_offsets(), &[0, 2, 3, 4]);
}
```

This currently panics at `core/src/reader.rs:793` with `index out of bounds: the len is 2 but the index is 2`. For projected ARRAY columns, the reader needs to fetch all physical columns required to reassemble the logical list (lengths plus child value/inner-length columns), and `column_readers` / `slot_locations` should be sized and typed by the expanded physical layout rather than by logical `global_indices`.

leaves12138 (approved):
Thanks for the update. I re-reviewed the latest head (`6e5968a`) and the previous correctness issues look fixed now.

I verified locally:
- `cargo test -p paimon-mosaic-core --test array_type_test`
- `cargo test -p paimon-mosaic-core --test robustness_test projection -- --nocapture`
- Extra temporary regressions for null-list child-range canonicalization and paged ARRAY projection both pass.

The current implementation canonicalizes null list rows on write by storing length 0 and skipping their child value ranges, and the paged projection path now expands logical ARRAY columns to the required physical slots. Looks good to me.

JingsongLi:
I found several correctness issues in the ARRAY implementation that should be fixed before merge:

1. ARRAY child columns are corrupted when the child column uses dictionary encoding.

`BucketWriter::write_dict_bit_packed` still iterates over `self.num_rows`, but for ARRAY child columns the physical row count is the number of elements, not the number of parent rows. A repro is 10 parent rows where each row contains 20 repeated `int32` elements such as `j % 3`: the child column chooses DICT, only the first 10 dictionary indexes are written, and the remaining decoded elements come back as dictionary index 0. Please iterate over `self.col_num_rows(col)` and use the child column's own null bitmap row count.

2. Multiple top-level ARRAY columns in the same bucket can be reassembled with the wrong lengths column.

`reassemble_list_columns` treats any child with `phys_idx > num_primary` as nested and uses `phys_idx - 1` as its lengths column. That is also true for the first child of the second top-level ARRAY column in a bucket, so `arr_b` can accidentally use `arr_a`'s child values as lengths. With `arr_a: list<int32>` and `arr_b: list<int64>` in one bucket, reading can panic while constructing a `ListArray` because the offsets exceed the values length. The fix should track the lengths physical index explicitly, or otherwise distinguish first-level children from nested children per parent.

3. ARRAYs with null parent rows lose data for several supported element types.

`validate_data_type` allows `List<Date32>`, `List<Time32>`, `List<Decimal128>`, `List<Timestamp>`, etc., and the normal write path supports those element arrays. But when a parent list has null rows, `flatten_list_values` calls `take_array`, whose fallback for unhandled types is `array.slice(0, 0)`. For example, `List<Date32>` with a null parent row that owns child slots reassembles with offsets pointing past an empty values array. Please either use Arrow's generic take implementation or handle every element type accepted by `validate_data_type`.

core/src/reader.rs
```diff
@@ -979,12 +1080,41 @@ impl RowGroupReader {
 
             match state {
                 BucketState::Paged { column_readers } => {
+                    // Read all physical columns (N+C)
+                    let mut phys_arrays: Vec<ArrayRef> = Vec::new();
+                    for cr_opt in column_readers {
+                        if let Some(ref cr) = cr_opt {
+                            phys_arrays.push(cr.read_all()?);
+                        } else {
+                            phys_arrays
```
QuakeWang: This still breaks paged partial projection when a bucket contains multiple ARRAY columns and only one of them is projected. Unprojected physical columns are materialized as Int32 null placeholders here, but the code below still calls reassemble_list_columns_pub for all ARRAY children in the bucket. For example, with arr_a: ARRAY<INT32> and arr_b: ARRAY<INT64> in the same paged bucket, projecting only arr_a tries to rebuild arr_b with an Int32 child placeholder and panics with "ListArray expected data type Int64 got Int32". Please either reassemble only projected ARRAY parents, or create placeholders using the expanded physical type and correct child row count. A regression test should cover multiple ARRAY columns + paged layout + projecting only one ARRAY.

docs/design.html
```diff
@@ -739,9 +739,10 @@ <h3>TypeDescriptor</h3>
                     <tr><td>15</td><td>TIME</td><td>varint precision</td></tr>
                     <tr><td>16</td><td>TIMESTAMP</td><td>varint precision</td></tr>
                     <tr><td>17</td><td>TIMESTAMP_LTZ</td><td>varint precision, varint timezoneLength, bytes timezone</td></tr>
+                    <tr><td>18</td><td>ARRAY</td><td>TypeDescriptor (recursive element type)</td></tr>
```
QuakeWang: This ARRAY type descriptor is stale after the schema roundtrip fix. The implementation now serializes the ARRAY element field name before the recursive element Field descriptor (`name length + name bytes + serialize_field(element_field)`), but the format spec still says only "TypeDescriptor". Please update this row so cross-language readers/writers know the exact on-disk schema encoding. Also note that the value serialization table below still mentions a "values sub-bucket", while the current layout stores ARRAY child values as physical columns in the same bucket.

JingsongLi:
I re-reviewed the latest head (`60ef99a`). The previous ARRAY DICT, multi-ARRAY reassembly, and Date32 null-parent issues look fixed, and the flattened ORC-style storage design itself looks reasonable: representing ARRAY as a primary INT32 lengths column plus child physical columns in the same bucket is simple, projection-friendly, and lets lengths/values reuse the existing CONST/DICT/PLAIN encoders.

I still found two issues that should be addressed:

1. The format version still stayed at 1 even though the bucket payload layout changed for all files.

`BucketWriter::finish()` now always prefixes monolithic bucket payloads with `varint(num_primary) + varint(num_children) + child counts`, even when the schema has no ARRAY columns. Paged buckets also now have the child-count header before the directory. However `spec::VERSION` remains `1`, so older v1 readers will accept new files and parse the new bucket header as old encoding flags / directory bytes, while the new reader will also try to parse old v1 files as if the new header existed.

This is not only an ARRAY compatibility issue; a new file with only primitive columns is no longer v1-compatible. Please bump the on-disk format version and either reject old v1 files cleanly or add a v1/v2 bucket parser split.

2. `ARRAY<legacy timestamp nanos struct>` is accepted by schema validation but can still panic on null parent rows.

`validate_data_type` accepts the legacy timestamp nanos struct (`Struct{millis: Int64, nanos_of_milli: Int32}`), and `DataType::List(field)` recursively accepts it as an ARRAY element type. The normal write path can downcast that struct element type, but when a list has null parent rows, `flatten_list_values()` calls `take_array()`. `take_array()` has no `Struct(fields) if is_timestamp_nanos_struct(fields)` branch and now panics in the fallback.

I reproduced this with a temporary test using `ListArray<Struct(millis, nanos_of_milli)>` with offsets `[0, 2, 4, 5]`, validity `[valid, null, valid]`, and struct child values. The write panics with:

```text
take_array: unsupported DataType Struct([Field { name: "millis", data_type: Int64 }, Field { name: "nanos_of_milli", data_type: Int32 }])
```

Please either reject legacy timestamp nanos structs when they appear inside ARRAY, or add a `take_array` branch for this accepted struct type and a regression test with a null parent list row.

leaves12138 (approved):
I re-reviewed the latest head (`7e6651e`) after the substantial follow-up changes. The issues raised after my previous approval are addressed now:

- ARRAY child DICT encoding now uses the physical child row count (`col_num_rows(col)`) when writing dictionary indexes.
- Multiple top-level ARRAY columns in the same bucket are covered, including the corrected `num_buckets = 1` regression.
- `take_array` now covers the supported ARRAY element types used when canonicalizing null parent rows, and `ARRAY<legacy timestamp nanos struct>` is rejected by validation.
- Non-ARRAY v1 compatibility is restored by omitting the new bucket header for buckets without ARRAY columns; I also verified the no-ARRAY golden against the actual base commit output.
- Paged multi-ARRAY projection and ARRAY golden tests pass.

Local validation I ran:
- `cargo test -p paimon-mosaic-core --test array_type_test -- --nocapture`
- `cargo test -p paimon-mosaic-core --test gen_fixtures -- --nocapture`
- `cargo test -p paimon-mosaic-core --test robustness_test corrupted -- --nocapture`
- `cargo test -p paimon-mosaic-core --test robustness_test projection -- --nocapture`
- `cargo test -p paimon-mosaic-core` (all passed until `interop_read_test`, which requires external `/tmp/mosaic_interop/*_written.mosaic` fixtures)
- Extra temporary regressions for nested ARRAY + sibling ARRAY in one bucket, and paged projection of a nested ARRAY from a multi-ARRAY bucket.

No further blocking issues from me.

QuakeWang (approved):
+1

PR #53, opened 2026-06-06
Support MAP type with flattened columnar storage

JingsongLi:
## Summary

Add MAP type support using the same ORC-style flattened columnar storage as ARRAY. A `MAP<K, V>` column is decomposed into three physical columns within the same bucket: lengths (INT32), keys (type K), and values (type V). All three independently benefit from DICT/CONST/PLAIN encoding.

Fully backward compatible — no version bump. MAP uses type byte 19. Older readers reject at schema validation.

## Storage Format

```
MAP<INT, UTF8> with rows: {1:"a", 2:"b"}, null, {3:"c"}

Lengths (INT32):  2, null, 1    ← DICT/CONST encoding
Keys    (INT32):  1, 2, 3       ← shared DICT across all maps
Values  (UTF8):   "a", "b", "c" ← shared DICT across all maps
```

## Changes

- **core/types.rs**: Type byte 19, validate, serialize/deserialize MAP
- **core/bucket_writer.rs**: `expand_col_types` for MAP (3 children), `extract_map_lengths`, `flatten_map_entries`
- **core/bucket_reader.rs**: `reassemble_map_array` from physical columns
- **docs/design.html**: MAP Type Storage section, TypeDescriptor table, Limitations update
- **docs/java-api.html, python-api.html, cpp-api.html**: MAP API examples
- **Golden file**: `v1_with_map.mosaic` binary compatibility test

## Test Plan

- [x] Rust: 3 MAP tests (basic, all-null, mixed with ARRAY) + golden file tests
- [x] Java: 1 MAP test (`testMapType`)
- [x] Python: 2 MAP tests (basic, null values)
- [x] C++: 1 MAP test
- [x] All existing ARRAY tests pass unchanged
- [x] Golden files: v1_no_array + v1_with_array unchanged


leaves12138 (changes requested):
Thanks for adding MAP support and the cross-language coverage. I checked the main flattened storage path (including null maps, all-null maps, MAP with other columns, projected MAP columns, multiple MAP columns in the same bucket, and null rows with hidden entry ranges), and those paths look good to me.

However, I found one blocker in the MAP schema serialization: the TypeDescriptor currently does not round-trip the full Arrow MAP metadata.

`serialize_field` writes only the two entry child fields for `DataType::Map`, while `deserialize_field` always reconstructs the type as `Map(Field("entries", Struct([Field("key", ...), Field("value", ...)]), false), false)`. This loses:

- the `sorted` flag from `DataType::Map(_, sorted)`;
- the map entries field name;
- the key/value field names.

That means a valid schema can be silently changed after write/read. The data values may still read correctly, but the Arrow schema no longer matches the original schema, and this is part of the file format metadata contract.

A minimal regression test that fails on this PR:

```rust
#[test]
fn test_review_map_field_names_and_sorted_roundtrip() {
    use arrow_buffer::{OffsetBuffer, ScalarBuffer};

    let key_field = Arc::new(Field::new("k_custom", DataType::Int32, false));
    let value_field = Arc::new(Field::new("v_custom", DataType::Utf8, true));
    let entries_field = Arc::new(Field::new(
        "my_entries",
        DataType::Struct(arrow_schema::Fields::from(vec![
            key_field.as_ref().clone(),
            value_field.as_ref().clone(),
        ])),
        false,
    ));
        "m",
        DataType::Map(entries_field.clone(), true),
        true,
    )]);

    let keys = Arc::new(Int32Array::from(vec![1])) as ArrayRef;
    let values = Arc::new(StringArray::from(vec![Some("a")])) as ArrayRef;
    let entries = StructArray::new(
        match entries_field.data_type() {
            DataType::Struct(fields) => fields.clone(),
            _ => unreachable!(),
        },
        vec![keys, values],
        None,
    );
    let map = MapArray::new(
        entries_field.clone(),
        OffsetBuffer::new(ScalarBuffer::from(vec![0i32, 1])),
        entries,
        None,
        true,
    );
    let batch = RecordBatch::try_new(Arc::new(schema.clone()), vec![Arc::new(map)]).unwrap();

    let result = roundtrip(&schema, &[batch]);
    let map_type = result[0].schema().field(0).data_type().clone();
    match map_type {
        DataType::Map(entries, sorted) => {
            assert!(sorted, "sorted flag should roundtrip");
            assert_eq!(entries.name(), "my_entries");
            if let DataType::Struct(fields) = entries.data_type() {
                assert_eq!(fields[0].name(), "k_custom");
                assert_eq!(fields[1].name(), "v_custom");
            } else {
                panic!("entries should be struct");
            }
        }
        other => panic!("expected map, got {:?}", other),
    }
}
```

The first failure is:

```text
sorted flag should roundtrip
```

Please extend the MAP TypeDescriptor payload to serialize/deserialize the entries field name and the `sorted` flag in addition to the key/value fields, and add a regression/golden test for this case. This should be fixed before merging because otherwise files written with MAP columns can lose schema metadata even when the value data is correct.

Local checks I ran:

- `cargo test -p paimon-mosaic-core --test array_type_test map -- --nocapture`
- `cargo test -p paimon-mosaic-core --test gen_fixtures map -- --nocapture`
- extra temporary tests for null hidden entries, multiple MAP columns in one bucket, projected MAP columns, and the failing schema metadata round-trip above

leaves12138 (changes requested):
Thanks for the update. The previous top-level MAP schema metadata issue is fixed for the supported unsorted MAP case: custom entries/key/value field names round-trip now, and sorted MAP is explicitly rejected during schema validation, which is a reasonable format boundary.

I found one remaining blocker around MAP keys with complex types.

`validate_data_type` currently accepts both MAP key and value types recursively. That means schemas such as `MAP<ARRAY<INT32>, UTF8>` pass `MosaicWriter::new`. However, the writer path does not actually support this shape: during `write_batch`, the top-level map split stores only `keys` and `values` in `ColSplit.child_arrays`, and then assigns them to the first two child descriptors for that parent. For a complex key, `expand_type` creates children for the key lengths and key leaf before the value child, so the second child descriptor is the key leaf, not the MAP value column. The result is a runtime write failure.

Minimal repro:

```rust
#[test]
fn test_map_with_array_key() {
    let key_type = DataType::List(Arc::new(Field::new("item", DataType::Int32, true)));
    let entries_field = Arc::new(Field::new(
        "entries",
        DataType::Struct(arrow_schema::Fields::from(vec![
            Field::new("keys", key_type, false),
            Field::new("values", DataType::Utf8, true),
        ])),
        false,
    ));
        "m",
        DataType::Map(entries_field.clone(), false),
        true,
    )]);

    let key_builder = ListBuilder::new(Int32Builder::new());
    let value_builder = StringBuilder::new();
    let mut map_builder = MapBuilder::new(None, key_builder, value_builder);
    map_builder.keys().values().append_value(1);
    map_builder.keys().append(true);
    map_builder.values().append_value("a");
    map_builder.append(true).unwrap();
    let batch = RecordBatch::try_new(
        Arc::new(schema.clone()),
        vec![Arc::new(map_builder.finish())],
    )
    .unwrap();

    let mut writer = MosaicWriter::new(out, &schema, WriterOptions::default()).unwrap();
    writer.write_batch(&batch).unwrap();
}
```

This fails with:

```text
Custom { kind: InvalidInput, error: "array downcast failed for DataType: Int32" }
```

Please either fully support complex MAP keys in the flatten/reassembly child mapping, or reject complex key types in `validate_data_type` with a clear error. Since Arrow MAP keys are typically scalar/non-null and the current implementation already supports complex MAP values, I think the simpler and safer fix is to reject `List`/`Map` key types for now and add a regression test that `MosaicWriter::new` fails early.

Local checks run on `f591b65c10ac2758598e139e58d3343dd95bf627`:

- `cargo test -p paimon-mosaic-core --test array_type_test map -- --nocapture`
- `cargo test -p paimon-mosaic-core --test gen_fixtures map -- --nocapture`
- temporary checks for custom MAP field names, sorted MAP rejection, and the complex-key failure above

leaves12138 (changes requested):
Thanks for the quick fix. I rechecked the latest head (`1d3dbb60d42cd048f822916f484adc38b5536c85`). The complex MAP key issue is fixed in the implementation by rejecting `List`/`Map` key types during schema validation, and my local regression for `MAP<ARRAY<INT32>, UTF8>` now fails early in `MosaicWriter::new` with the expected error. I also rechecked that custom top-level MAP entries/key/value field names round-trip correctly for unsorted MAP.

I still want to keep this as changes-requested until the regression tests are added to the PR. The fixes cover subtle file-format/schema-contract bugs, but the current test suite does not appear to include:

- custom MAP entries/key/value field name round-trip;
- sorted MAP rejection;
- complex MAP key rejection, e.g. `MAP<ARRAY<INT32>, UTF8>` should fail in `MosaicWriter::new`.

Please add focused tests for these validation/schema cases. After that I think this PR should be good to merge from my side.

Local checks I ran:

- `cargo test -p paimon-mosaic-core --test array_type_test map -- --nocapture`
- `cargo test -p paimon-mosaic-core --test gen_fixtures map -- --nocapture`
- temporary regression tests for custom MAP field names and complex MAP key rejection

leaves12138 (approved):
Thanks for the updates. I rechecked the latest head (`a6710d09eb6bbd4a4d25fb8a8d1256ce00b19e43`). The previous issues are now covered:

- custom MAP entries/key/value field names round-trip correctly;
- sorted MAP is explicitly rejected;
- complex MAP keys are rejected early during schema validation;
- the focused regression tests are included in `array_type_test.rs`.

I also re-ran the targeted local checks:

- `cargo test -p paimon-mosaic-core --test array_type_test map -- --nocapture`
- `cargo test -p paimon-mosaic-core --test gen_fixtures map -- --nocapture`

The GitHub checks are green as well. Looks good to me now.

PR #54, opened 2026-06-07
[WIP] support STRUCT type with column ID and Schema-based projection

JingsongLi:
### STRUCT Storage
- Top-level STRUCT fields expanded to independent columns across buckets for projection pushdown
- DFS-incremented column IDs (like ORC) for all internal column identity
- Physical column names use dot-prefixed format (`info.name`) for bucket distribution only; all matching uses column IDs
- Synthetic `__null__` columns tracked by `null_col_ids` — no name-based detection
- Schema serializes original STRUCT types (type byte 20); reader forward-expands on deserialize

### Nested Types
- ARRAY<STRUCT> and MAP<K, STRUCT> with ORC-style child columns within bucket
- Recursive nested STRUCT support: `ARRAY<STRUCT<name, addr: STRUCT<city, zip>>>` correctly reassembled
- `__struct_null__` child carries original STRUCT data type for recursive reassembly
- Per-level `__null__` columns for nested nullable STRUCT null semantics
- STRUCT inside MAP key rejected at validation

### Projection (Schema-based, like Lance)
- `project_schema(&Schema)` as the primary projection API — accepts Arrow Schema subset
- Partial STRUCT projection expressed naturally as Schema subsets, no dot-path parsing
- Column names and STRUCT field names can contain any characters including `.`
- Ancestor `__null__` columns auto-included in physical reads
- `row_group_reader_by_names` kept as convenience for top-level columns
- FFI accepts Arrow C Schema pointer; C++/Python/Java updated

### Correctness
- Non-nullable STRUCT fields/parents validated with ancestor null mask propagation
- Sliced StructArray null propagation handles bitmap offsets correctly
- Stats collection uses expanded arrays; ambiguous stats selections rejected
- `propagate_struct_nulls` shared as crate-level function

### Documentation
- Complex types design in standalone `docs/complex-types.html`

## Test plan

- [x] STRUCT basic/nullable/nested roundtrip with field order preservation
- [x] Nested nullable STRUCT null semantics (addr=null vs addr={city:null, zip:null})
- [x] Nested STRUCT null preserved in leaf projection
- [x] Schema-based leaf, whole-STRUCT, nested, and mixed projection
- [x] ARRAY<STRUCT> and MAP<K, STRUCT> roundtrip
- [x] ARRAY<STRUCT<nested STRUCT>> roundtrip
- [x] STRUCT leaf stats with null propagation
- [x] Non-nullable STRUCT field/parent rejects null batches (with ancestor mask)
- [x] Sliced StructArray null propagation
- [x] Column names containing `.` roundtrip and coexist with STRUCT
- [x] Top-level column and STRUCT field with same name
- [x] STRUCT field names containing `.` allowed and roundtrip
- [x] MAP<STRUCT key> rejected
- [x] Python/Java/C++ Schema-based projection
- [x] Golden file stable
- [x] 125 lib + 48 array_type + 8 golden + 30 encoding + 53 format + 26 robustness + 41 Python tests pass
- [x] Workspace clippy clean, fmt clean


leaves12138 (changes requested):
Thanks for adding top-level STRUCT expansion. I reviewed the Rust writer/reader/schema paths and ran:

- `cargo test -p paimon-mosaic-core --test array_type_test test_struct -- --nocapture`
- `cargo test --workspace --all-targets -- --nocapture`

The STRUCT tests pass. The full workspace test run gets through the core tests but fails at `core/tests/interop_read_test.rs` because `/tmp/mosaic_interop/{java,python}_written.mosaic` is not present in this environment.

I found one correctness blocker around STRUCT nullability validation. When expanding STRUCT fields, `expand_struct_fields_recursive` emits every leaf column as nullable (`out.push((full_name, dt.clone(), true))`). Then `write_batch` validates non-nullability only against these expanded physical columns. This means a writer schema can declare `info.age` as non-nullable, but a batch whose `info.age` array is nullable and contains nulls is accepted. The file is written successfully, but reading it back panics when the reader reconstructs the original `StructArray` with the non-nullable `age` field.

Minimal regression:

```rust
#[test]
fn test_non_nullable_struct_leaf_rejects_nulls_from_batch_schema() {
    let writer_schema = Schema::new(vec![Field::new(
        "info",
        DataType::Struct(arrow_schema::Fields::from(vec![Field::new(
            "age",
            DataType::Int32,
            false,
        )])),
        true,
    )]);
    let batch_schema = Schema::new(vec![Field::new(
        "info",
        DataType::Struct(arrow_schema::Fields::from(vec![Field::new(
            "age",
            DataType::Int32,
            true,
        )])),
        true,
    )]);
    let ages = Int32Array::from(vec![Some(1), None]);
    let info = StructArray::new(
        arrow_schema::Fields::from(vec![Field::new("age", DataType::Int32, true)]),
        vec![Arc::new(ages) as ArrayRef],
        None,
    );
    let batch = RecordBatch::try_new(Arc::new(batch_schema), vec![Arc::new(info)]).unwrap();

    let mut writer = MosaicWriter::new(out, &writer_schema, WriterOptions::default()).unwrap();
    assert!(writer.write_batch(&batch).is_err());
}
```

A related case also slips through: a writer schema can declare the parent `info` STRUCT as non-nullable, while the incoming batch schema marks it nullable and contains null parent STRUCT values. Because non-nullable top-level STRUCTs do not get an `info.__null__` expanded column, this parent nullability is not checked either.

Please preserve each expanded leaf's declared `Field::is_nullable()` and add explicit validation for non-nullable STRUCT parents/ancestors before writing. It would also be good to add regression tests for both non-nullable STRUCT leaf fields and non-nullable parent STRUCTs, including a readback check that a file accepted by the writer cannot later panic during `StructArray` reconstruction.

leaves12138 (changes requested):
Thanks for the update. The previous non-nullable STRUCT parent/leaf validation issue is fixed: the added `test_non_nullable_struct_leaf_rejects_nulls` and `test_non_nullable_struct_parent_rejects_nulls` both pass locally.

I found another correctness issue in the same null-propagation path. `propagate_struct_nulls` copies the raw null bitmap bytes from `child.nulls().inner().inner()` and `struct_arr.nulls().inner().inner()` directly into a new bitmap starting at offset 0. This ignores Arrow array offsets. If the incoming `StructArray` (or one of its child arrays) is a slice with a non-zero offset, the propagated child validity is shifted and valid rows can be written as null (or vice versa). The same byte-copy pattern also exists in `bucket_writer::propagate_struct_nulls`, so ARRAY/MAP<STRUCT> elements have the same risk.

Minimal regression for the top-level STRUCT path:

```rust
#[test]
fn test_sliced_struct_child_nulls_are_offset_correct() {
        "info",
        DataType::Struct(arrow_schema::Fields::from(vec![Field::new(
            "name",
            DataType::Utf8,
            true,
        )])),
        true,
    )]);
    let names = StringArray::from(vec![Some("skip"), Some("a"), Some("b"), Some("c")]);
    // Original validity: row0 valid, row1 null, row2 valid, row3 null.
    let null_buf = NullBuffer::new(BooleanBuffer::new(Buffer::from(vec![0b0000_0101]), 0, 4));
    let info = StructArray::new(
        arrow_schema::Fields::from(vec![Field::new("name", DataType::Utf8, true)]),
        vec![Arc::new(names) as ArrayRef],
        Some(null_buf),
    );
    let sliced = Arc::new(info.slice(1, 3)) as ArrayRef;
    let batch = RecordBatch::try_new(Arc::new(schema.clone()), vec![sliced]).unwrap();

    let result = roundtrip_projected(&schema, &[batch], &["info.name"]);
    let out = result[0].column(0).as_any().downcast_ref::<StructArray>().unwrap();
    assert!(out.is_null(0));
    assert!(!out.is_null(1));
    assert!(out.is_null(2));
    let name_out = out.column_by_name("name").unwrap().as_any().downcast_ref::<StringArray>().unwrap();
    assert!(!name_out.is_null(1), "child field under valid sliced parent must not be null");
    assert_eq!(name_out.value(1), "b");
}
```

This fails on the latest head: `name_out.is_null(1)` is true, even though row 1 in the sliced batch corresponds to the valid original row 2. Please build the propagated bitmap by iterating logical row positions (`child.is_valid(i) && struct_arr.is_valid(i)`), or otherwise account for `BooleanBuffer` offsets instead of copying the underlying bytes. Please fix both `writer.rs` and `bucket_writer.rs` and add sliced-array regression tests for top-level STRUCT and STRUCT inside ARRAY/MAP.

leaves12138 (changes requested):
Thanks for the quick fix. I re-reviewed the latest head `cf4f0c30f050530cf2b688c92c6e329e496b227a`.

The previous sliced `StructArray` bitmap issue is fixed in both `writer.rs` and `bucket_writer.rs`: the new code builds the propagated bitmap by iterating logical row positions, so it correctly accounts for Arrow offsets. I verified the existing regressions and also tried extra local cases where the sliced child has its own null bitmap and where parent + child nulls are combined; those pass.

Commands run on a clean checkout of this head:

- `cargo test -p paimon-mosaic-core --test array_type_test test_struct -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_array_of_struct -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_map_of_struct_value -- --nocapture`

I found one remaining blocker in the Python API wrapper for the newly documented STRUCT leaf projection mode. `python/mosaic/mosaic.py::MosaicReader.project()` calls the native projection first, then unconditionally builds `_projected_schema` with `self._schema.field(name)` for each requested name:

```python
projected_field_names = list(dict.fromkeys(column_names))
self._projected_schema = pa.schema(
    [self._schema.field(name) for name in projected_field_names],
    metadata=self._schema.metadata,
)
```

This works for top-level columns like `"info"`, but it fails for documented STRUCT leaf projections such as `"info.name"`: PyArrow schemas do not resolve dotted nested field paths through `schema.field("info.name")`. So `reader.project(["info.name"])` can succeed in Rust/native, but then the Python wrapper raises `KeyError: 'Column info.name does not exist in schema'` before users can read anything.

Minimal Python-side repro for the wrapper failure:

```python
import pyarrow as pa

schema = pa.schema([
    pa.field("info", pa.struct([
        pa.field("name", pa.utf8()),
        pa.field("age", pa.int32()),
    ]))
])

schema.field("info")       # ok
schema.field("info.name")  # KeyError: Column info.name does not exist in schema
```

This is user-visible because the PR documents leaf projection as `project(&["info.name"])`, and Python exposes the same `project(columns)` API. Please update the Python wrapper to derive projected schemas for STRUCT leaf projections (for example, by asking native/Rust for the projected output schema, or by implementing dotted-path reconstruction that returns a partial top-level `info` struct), and add a Python regression test for `reader.project(["info.name"])` on a STRUCT column.

There is also a related Rust API gap worth covering: `resolve_projection` currently accepts a top-level STRUCT name (`"info"`) and leaf names (`"info.name"`), but not a nested STRUCT field name itself (`"info.addr"`) even though nested leaves like `"info.addr.city"` exist. If the intended projection contract is only whole top-level STRUCT plus leaves, this should be clarified in docs; otherwise please support whole nested-STRUCT field projection too.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed latest head `7da887ccef30e65f09b2caa54afe7f339e6fb7e2`.

The previous Python top-level STRUCT leaf projection failure is mostly addressed: `reader.project(["info.name"])` now has a Python regression, and Rust now also accepts the nested STRUCT path projection `"info.addr"`. I verified the Rust side on a clean checkout with:

- `cargo test -p paimon-mosaic-core --test array_type_test test_struct -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_struct_nested_path_projection -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_array_of_struct -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_map_of_struct_value -- --nocapture`

I found one remaining Python wrapper correctness issue in the new `_build_projected_schema` helper. It deduplicates at only the first child name under each top-level STRUCT:

```python
child_name = leaf_parts[0]
if child_name in child_names_seen:
    continue
child_names_seen.add(child_name)
child = _resolve_field(top_field.type, leaf_parts)
```

This drops later projected leaves that share the same nested STRUCT parent. For example, with schema `info: struct<addr: struct<city: string, zip: int32>>`, projecting both `info.addr.zip` and `info.addr.city` should produce a partial schema containing both `addr.zip` and `addr.city`. The Rust reader does return both fields (I added a local Rust regression for `roundtrip_projected(..., &["info.addr.zip", "info.addr.city"])`, and it passed), but the Python `_projected_schema` for empty-result paths only keeps the first one:

```python
schema = pa.schema([
    pa.field("info", pa.struct([
        pa.field("addr", pa.struct([
            pa.field("city", pa.utf8()),
            pa.field("zip", pa.int32()),
        ])),
    ])),
])

_build_projected_schema(schema, ["info.addr.zip", "info.addr.city"])
# currently: info: struct<addr: struct<zip: int32>>
# expected:  info: struct<addr: struct<zip: int32, city: string>>
```

This can make `read_all()` on an empty projected file report a schema that is narrower than the native projection contract, and it is inconsistent with non-empty reads. Please make the Python projected-schema reconstruction merge nested dotted paths recursively instead of deduplicating only by the immediate child name, and add a Python regression covering two leaves under the same nested STRUCT parent, e.g. `reader.project(["info.addr.zip", "info.addr.city"])` with a zero-row or otherwise empty-result read.

leaves12138 (approved):
Thanks for the update. I re-reviewed latest head `eeccc287d7580dde01ae54fdcffc58825922f4c2`.

The previously reported issues are fixed:
- STRUCT parent/leaf nullability validation is covered.
- Sliced `StructArray` null propagation now iterates logical rows and handles Arrow offsets.
- Python `reader.project(["info.name"])` no longer fails on dotted paths.
- The Python projected-schema helper now recursively merges nested dotted paths, so `reader.project(["info.addr.zip", "info.addr.city"])` keeps both leaves.
- Rust supports nested STRUCT path projection such as `info.addr`.

Validation:
- `cargo test -p paimon-mosaic-core --test array_type_test test_struct -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_struct_nested_path_projection -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_array_of_struct -- --nocapture`
- `cargo test -p paimon-mosaic-core --test array_type_test test_map_of_struct_value -- --nocapture`
- `cargo build -p paimon-mosaic-ffi`
- `MOSAIC_LIB_PATH=target/debug python -m pytest python/tests/test_mosaic.py -k 'struct' -q`

I do not have further blocking comments.

core/src/bucket_reader.rs
```diff
@@ -393,16 +393,184 @@ fn reassemble_list_columns(
...
+            });
+        if is_struct_field_child {
+            continue;
+        }
+
+        // Check if this is a __struct_null__ child — if so, reassemble the STRUCT
+        if child.element_field.name() == "__struct_null__" {
+            // Collect all subsequent children with same parent that aren't processed
+            let mut struct_field_indices: Vec<usize> = Vec::new();
+            for si in (idx + 1)..children.len() {
```
QuakeWang: Collecting “all following unprocessed children with the same parent” works for a flat `ARRAY<STRUCT<a, b>>`, but it flattens nested STRUCT fields.

For `ARRAY<STRUCT<name, addr: STRUCT<city, zip>>>`, the child layout is roughly:

```rust
__struct_null__, name, addr.__struct_null__, city, zip
```

This code collects `addr.__struct_null__`, `city`, and `zip` as fields of the outer STRUCT, so readback becomes `STRUCT<name, __struct_null__, city, zip>` instead of `STRUCT<name, addr: STRUCT<city, zip>>`, and `RecordBatch::try_new` fails with a schema mismatch.

I think this should rebuild STRUCT elements by recursively consuming child columns from the original `DataType::Struct`, and add regressions for `ARRAY<STRUCT<nested STRUCT>>` and `MAP<K, STRUCT<nested STRUCT>>`.

ffi/src/lib.rs
QuakeWang, line 582: This still exports the internal expanded schema. For files containing a `STRUCT`, `schema.columns` includes physical columns such as `info.__null__`, `info.name`, and `info.age`, but the reader’s public schema should match the logical Arrow schema that the user wrote.

A small `id`, `info: struct<name, age>` case currently gives:

```rust
reader.schema.names = ['id', 'info.__null__', 'info.name', 'info.age']
read_row_group(0).schema.names = ['id', 'info']
```

So the Python/C++ visible `reader.schema` does not match the schema of the batches returned by reads. I think this should use the original logical columns when `original_columns` is present, instead of reconstructing the public schema from expanded `schema.columns`.

core/src/schema.rs
```diff
@@ -62,10 +96,328 @@ impl MosaicSchema {
...
+                        }
+                    }
+                } else {
+                    // Non-STRUCT field matching a STRUCT column — expand all
+                    output.extend_from_slice(&mapping.expanded_col_indices);
+                }
+            } else {
+                // Non-STRUCT column: find by name in original schema
+                let orig = self.original_columns.as_deref();
+                let found =
```
QuakeWang: This checks that the projected field exists in `original_columns`, but then still resolves it by first matching `ColumnMeta.name` in the expanded columns. That name is not unique in this format: a top-level column `"a.b"` and the leaf `b` inside `STRUCT a` can both produce the physical name `"a.b"`.

A failing case is:

```rust

  Schema::new(vec![
      Field::new("a", DataType::Struct(... field "b" ...), true),
      Field::new("a.b", DataType::Int32, false),
  ])
```

With projection `Schema::new(vec![Field::new("a.b", DataType::Int32, false)])`, the current lookup can pick the `STRUCT a` leaf and the output field becomes `"a"` instead of the top-level `"a.b"`. This should resolve top-level non-STRUCT fields through their original schema position / column id, not by first matching the physical column name.

core/src/bucket_reader.rs
```diff
@@ -393,16 +480,167 @@ fn reassemble_list_columns(
...
+                    key_child.physical_index - 1
+                } else {
+                    parent
+                };
+
+                let lengths = arrays[lengths_idx].clone();
+                let lengths_rows = lengths.len();
+
+                let key_field = key_child.element_field.as_ref().clone();
+                let val_field = Field::new(
```
QuakeWang, line 580: This rebuilds the `MAP<K, STRUCT<...>>` entries/value fields with hard-coded `"entries"` and `"values"` names. That loses the original Arrow field names from the schema.

The project already has `test_map_custom_field_names_roundtrip` for preserving custom map field names in the normal MAP path. When the value is a `STRUCT`, this new branch changes a schema like:

```
Map("my_entries": Struct("k_custom", "v_custom": Struct(...)))
```

into:

```
  Map("entries": Struct("k_custom", "values": Struct(...)))
```

In a local repro, this makes `RecordBatch::try_new` fail with a schema type mismatch. I think this branch should take the original entries_field and value field name from the map container type, and only replace/update the value `Struct` data type and array instead of hard-coding new field names.

core/src/bucket_reader.rs
```diff
@@ -393,16 +480,167 @@ fn reassemble_list_columns(
...
+                });
+                let lengths_idx = if prev_child_is_lengths {
+                    phys_idx - 1
+                } else {
+                    parent
+                };
+
+                let lengths = arrays[lengths_idx].clone();
+                let lengths_rows = lengths.len();
+                let element_field = Arc::new(Field::new(
```
QuakeWang, line 617: The same issue applies to `ARRAY<STRUCT<...>>`. This hard-codes the list element field name to `"item"`, but Arrow’s list element field name is part of the schema, and this project also serializes/preserves it.

For example, with:

```
DataType::List(Arc::new(Field::new("element", struct_type, true)))
```

the read path reconstructs the array with the default `"item"` field name, which no longer matches the reader schema’s `"element"` field. That causes `RecordBatch::try_new` to fail with a type mismatch. This should copy the original list element field name/nullability and only replace the `DataType::Struct(...)` part.

JingsongLi:
At present, it is too complex, and we may need to break down the design to ensure that each step is solid.

PR #55, opened 2026-06-08
Refactor complex child column layout metadata

JingsongLi:
## Summary

Make ARRAY/MAP physical child layout explicit in the in-memory metadata used by bucket read/write paths. This keeps the on-disk format unchanged while removing order-based inference for nested complex child columns.

## Changes

- Add `ChildColumnRole` plus `length_physical_index` to `ChildColumnMeta` so child columns know whether they are list elements, map keys, or map values, and which physical lengths column owns them.
- Refactor complex type expansion and writer traversal to locate nested children through explicit layout metadata instead of sibling index offsets.
- Refactor reader reassembly to rebuild ARRAY/MAP columns from explicit roles and length ownership rather than adjacent-child heuristics.
- Add a unit test covering the physical layout plan for a nested `ARRAY<MAP<INT, ARRAY<UTF8>>>` shape.

## Testing

- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] `cargo test -p paimon-mosaic-core --lib`
- [x] `cargo test -p paimon-mosaic-core --test array_type_test`
- [x] `cargo test -p paimon-mosaic-core --test gen_fixtures`

## Notes

`cargo test -p paimon-mosaic-core` was also run. It passed through unit, ARRAY/MAP, encoding, format, and fixture tests, then failed in `interop_read_test` because `/tmp/mosaic_interop/python_written.mosaic` was not present; that test expects externally generated interop fixtures.

### DIFF
--- a/core/src/bucket_reader.rs
+++ b/core/src/bucket_reader.rs
@@ -368,135 +368,81 @@ fn reassemble_list_columns(
             continue;
         }
         let child = &children[idx];
-        let parent = child.parent_logical_col;
-        let phys_idx = child.physical_index;
-        let elem_dt = child.element_field.data_type();
+        match child.role {
+            ChildColumnRole::MapValue => {
+                let key_idx = children.iter().position(|candidate| {
+                    candidate.parent_logical_col == child.parent_logical_col
+                        && candidate.length_physical_index == child.length_physical_index
+                        && candidate.role == ChildColumnRole::MapKey
+                });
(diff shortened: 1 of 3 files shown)


QuakeWang (approved):
LGTM

Non-blocking: the new test mainly verifies the metadata layout. It would be useful to add a real roundtrip case such as `List<Map<Int32, List<Utf8>>>`, covering the writer pending queue and reader reassembly together.

PR #57, opened 2026-06-11
ci: fix update_branch_version.sh skipping Cargo.toml on snapshot bumps

jianguotian:
update_branch_version.sh bumps the version across the pom, the three Cargo.toml crates, and pyproject.toml. The pom regex tolerates an optional -SNAPSHOT suffix, but Cargo.toml and pyproject.toml are matched against OLD_VERSION verbatim — and those files never carry -SNAPSHOT.

So when OLD_VERSION is 0.2.0-SNAPSHOT (the normal "bump main to 0.3.0-SNAPSHOT" case), the perl match fails on the Rust crates and the Python package: they stay at the old version while the pom moves. main is already in this split state today (pom 0.2.0-SNAPSHOT, crates 0.2.0).

Strip -SNAPSHOT from OLD_VERSION too, mirroring NEW_VERSION_CLEAN. After this, OLD=0.2.0-SNAPSHOT NEW=0.3.0-SNAPSHOT moves all five files in one run — verified locally: pom → 0.3.0-SNAPSHOT, crates + pyproject → 0.3.0.

### DIFF
--- a/tools/update_branch_version.sh
+++ b/tools/update_branch_version.sh
@@ -52,17 +52,19 @@ fi
 
 cd ..
 
-# For Cargo.toml and pyproject.toml, strip any -SNAPSHOT suffix (not valid in those ecosystems)
+# Cargo.toml and pyproject.toml never carry the -SNAPSHOT suffix, so strip it
+# from both old and new versions when matching/replacing there.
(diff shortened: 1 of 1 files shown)


JingsongLi:
+1

PR #58, opened 2026-06-11
chore: update dependency list for release 0.2.0

jianguotian:
DEPENDENCIES.rust.tsv records the third-party Rust crates bundled into the release and their SPDX licenses; ASF release policy requires it to match the actual dependency tree.

Cargo.lock is gitignored (library convention), so cargo resolves the latest semver-compatible versions on each build. Since 0.1.0 several crates drifted (serde_json 1.0.149→1.0.150, wasm-bindgen 0.2.121→0.2.123, chrono, zerocopy, etc.) and a few new transitive crates appeared (wit-bindgen, wasip3, r-efi), so the checked-in tsv no longer reflects the real set. The workspace crates also moved 0.1.0→0.2.0.

Regenerated with `python3 tools/dependencies.py generate`; `check` passes (licenses ok). The four tsv files now match the current tree, keeping the 0.2.0 source release license-compliant.

### DIFF
--- a/DEPENDENCIES.rust.tsv
+++ b/DEPENDENCIES.rust.tsv
@@ -10,15 +10,15 @@ arrow-array@58.3.0	X					X
 arrow-buffer@58.3.0	X								
 arrow-data@58.3.0	X								
 arrow-schema@58.3.0	X								
-autocfg@1.5.0	X					X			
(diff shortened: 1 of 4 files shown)


XiaoHongbo-Hope:
+1

PR #64, opened 2026-06-15
ci: build linux wheels in manylinux container, drop cibuildwheel

jianguotian:
**Problem:** the Release workflow will startup_failure on release tags — ASF removed `pypa/cibuildwheel` from the org Actions allowlist. The 0.2.0-rc1 wheels were built before the change, but cutting the final v0.2.0 tag would now fail (same issue already hit paimon-vector-index rc3).

**Fix:** build linux wheels inside the `manylinux_2_28` image. Container images aren't subject to the allowlist, so the python release path no longer depends on any third-party action (publish already uses allowlisted gh-action-pypi-publish). manylinux_2_28 / cp39 / ffi.so / auditwheel preserved; mac/win unchanged.

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -37,6 +37,10 @@ permissions:
 jobs:
   wheels-linux:
     runs-on: ${{ matrix.os }}
+    # Build inside the manylinux image directly. Container images are not subject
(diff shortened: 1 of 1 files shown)


jianguotian:
Closing for now per request — will revisit before the final 0.2.0 release.

PR #65, opened 2026-06-15
ci: bump cibuildwheel to v4.1.0 (allowlisted SHA)

jianguotian:
**DO NOT MERGE until 0.2.0 vote passes** — hold so the voted source tree is untouched; merge right before cutting the final v0.2.0 tag.

**Problem:** release-python.yml pins cibuildwheel at a 5-month-old v3.3.1 SHA. ASF tightened the Actions allowlist and dropped that SHA, so Release startup_failures on tags. 0.2.0-rc1 wheels predate the change, so the vote is unaffected.

**Fix:** bump to v4.1.0, a current allowlisted SHA. One line; mac/win unchanged.

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -62,7 +62,7 @@ jobs:
           python-version: "3.12"
 
       - name: Build wheels via cibuildwheel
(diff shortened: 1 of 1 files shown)


XiaoHongbo-Hope:
+1

PR #66, opened 2026-06-16
[cli] Add mosaic command-line tool

jianguotian:
Inspecting a Mosaic file used to mean writing Rust against the library API. This adds a mosaic binary to inspect, query and import files from the shell, with a command surface aligned to parquet-cli so the workflow is familiar.

## Commands — all support `--json`

| Command | Shows |
|---|---|
| schema | column names, Arrow types, nullability, bucket assignment |
| meta | row groups, row counts, per-column stats (null_count/min/max) |
| cat / head | rows as a table; `-n`, `--all`, `-c/--columns`, `--where` (stats pushdown) |
| count | total row count |
| convert | import CSV or JSON lines into a new file (schema inferred, `--stats`) |
| pages | per-column encoding (plain/const/dict/all_null) + slot size |
| footer | magic, version, bucket count, compression |
| column-size | on-disk bytes per column + total compression ratio |
| dictionary | dump a dict-encoded column (`-c/--column`) |
| buckets | bucket layout per row group (monolithic vs paged) + ratio |

```
mosaic convert data.csv -o data.mosaic --stats id
mosaic cat data.mosaic --all --where "id>100"   # skips row groups via min/max
mosaic count data.mosaic
```

## Structure

- New cli workspace crate: clap commands, text/JSON renderers, where filter + stats pushdown, CSV/JSON import; reads via a file-backed InputFile (pread)
- Core: read-only additive accessors only — no format or behavior change; convert uses the existing MosaicWriter
- Docs: docs/cli.html and cli/README.md cover every command with text/JSON examples

## Tests

221 pass, 0 fail — cli e2e covers all commands incl. compression ratio, where filter, stats pushdown boundaries and CSV/JSON round-trip; core unchanged (126 + 53 + 21)


JingsongLi:
Can you also add documentation? And can you compare the API design differences with Parquet CLI?

cli/src/fmt.rs
```diff
@@ -0,0 +1,321 @@
...
+        return "null".to_string();
+    }
+    use arrow_array::{Float32Array, Float64Array};
+    match arr.data_type() {
+        Utf8 => json_str(&cell(arr, row)),
+        // NaN / Infinity are not valid JSON numbers — emit null.
+        Float32 if !arr.as_any().downcast_ref::<Float32Array>().unwrap().value(row).is_finite() => "null".into(),
+        Float64 if !arr.as_any().downcast_ref::<Float64Array>().unwrap().value(row).is_finite() => "null".into(),
+        // Date32 is an epoch-day integer; emit bare like other numerics.
+        _ => cell(arr, row),
```
JingsongLi: `cat --json` can emit invalid JSON for Mosaic types that are supported by the core reader but not rendered by `cell()`. For example Binary, Decimal, Time32, Timestamp, List, and Map currently fall through to `?`, and this branch writes that value unquoted, producing output like `"col":?`. Parquet/Arrow CLI paths use a real JSON serializer or type-aware value rendering; please either route this through Arrow JSON writing or cover all supported Mosaic types explicitly.

cli/src/fmt.rs
```diff
@@ -0,0 +1,321 @@
...
+        use arrow_schema::DataType::*;
+        let numeric = matches!(col.data_type(), Int8|Int16|Int32|Int64|Float32|Float64|Date32);
+        if !numeric || w.value.parse::<f64>().is_err() {
+            return Err(format!("--where: '{}' needs a numeric column and value (got '{}' {} '{}')", w.op, w.column, w.op, w.value));
+        }
+    }
+    let mask: Vec<bool> = (0..batch.num_rows()).map(|r| {
+        if col.is_null(r) { return false; }
+        let lhs = cell(col.as_ref(), r);
+        match (lhs.parse::<f64>(), w.value.parse::<f64>()) {
```
JingsongLi: This makes equality semantics depend on whether the rendered strings happen to parse as numbers. On a Utf8 column, `--where s=1` matches both `"1"` and `"01"` because both sides parse as `f64`. Since Parquet CLI does not define this filter behavior, Mosaic should make the semantics type-driven: exact string comparison for Utf8, numeric comparison only for numeric columns.

cli/src/main.rs
```diff
@@ -0,0 +1,479 @@
...
+    let reader = open(file)?;
+    let s = reader.schema();
+    let mut bytes = vec![0usize; s.columns.len()];
+    let mut raw = vec![0usize; s.columns.len()];
+    for rg in 0..reader.num_row_groups() {
+        // On-disk bytes are tracked per bucket (works for both monolithic and
+        // paged layouts); split a bucket's size across its member columns.
+        for b in reader.bucket_infos(rg)? {
+            if b.columns.is_empty() { continue; }
+            split_evenly(b.size, &b.columns, &mut bytes);
```
JingsongLi, line 682: This differs materially from Parquet CLI `column-size`, which sums exact column-chunk sizes from metadata. Here the bucket size is split evenly across member columns, so the per-column bytes can be quite misleading; for paged buckets we should be able to use per-column slot sizes instead. If monolithic bucket attribution must remain approximate, please label it as approximate in the command output/docs rather than presenting it as exact per-column size.

cli/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```bash
+cargo run -p paimon-mosaic-cli -- <command> <file>   # run from source
+cargo install --path cli                             # install `mosaic`
+mosaic <command> <file>
+```
+
+## Commands
+
+Every command accepts `--json`.
```
JingsongLi: `convert` does not accept `--json`, so this statement is currently inaccurate. Please scope this to the inspection/query commands or add a JSON mode to `convert`.

JingsongLi (commented):
I left a few focused comments on CLI semantics and Parquet CLI alignment.

cli/src/fmt.rs
jianguotian: > `cat --json` can emit invalid JSON for Mosaic types that are supported by the core reader but not rendered by `cell()`. For example Binary, Decimal, Time32, Timestamp, List, and Map currently fall through to `?`, and this branch writes that value unquoted, producing output like `"col":?`. Parquet/Arrow CLI paths use a real JSON serializer or type-aware value rendering; please either route this through Arrow JSON writing or cover all supported Mosaic types explicitly.

Fixed, `cat --json` now goes through Arrow's JSON writer, so all reader types render as valid JSON.

cli/src/fmt.rs
jianguotian: > This makes equality semantics depend on whether the rendered strings happen to parse as numbers. On a Utf8 column, `--where s=1` matches both `"1"` and `"01"` because both sides parse as `f64`. Since Parquet CLI does not define this filter behavior, Mosaic should make the semantics type-driven: exact string comparison for Utf8, numeric comparison only for numeric columns.

Fixed, comparison is type-driven now: numeric columns compare numerically, others as exact strings. `s=01` no longer matches `1`.

cli/src/main.rs
jianguotian: > This differs materially from Parquet CLI `column-size`, which sums exact column-chunk sizes from metadata. Here the bucket size is split evenly across member columns, so the per-column bytes can be quite misleading; for paged buckets we should be able to use per-column slot sizes instead. If monolithic bucket attribution must remain approximate, please label it as approximate in the command output/docs rather than presenting it as exact per-column size.

Fixed, paged buckets use exact per-column slot sizes; monolithic multi-column buckets are split and labeled `(approx)`.

cli/README.md
jianguotian: > `convert` does not accept `--json`, so this statement is currently inaccurate. Please scope this to the inspection/query commands or add a JSON mode to `convert`.

Fixed, scoped the statement to inspection/query commands; `convert` writes a file.

jianguotian:
> Can you also add documentation? And can you compare the API design differences with Parquet CLI?

Docs added — `docs/cli.html` and `cli/README.md`.

Compared with parquet-cli:
1. Most commands is familiar with parquet-cli (`schema`/`meta`/`cat`/`count`/`convert`,
`-c`/`--json`/`--where`).
2. The new part is Mosaic's buckets: the `buckets` command shows the monolithic/paged
layout, and `schema`/`pages` show which bucket each column lands in:

```
$ mosaic buckets data.mosaic
    bucket 0: monolithic 15B [flag]
    bucket 1: paged 353B [id]
    bucket 2: paged 32B [kind]

$ mosaic schema data.mosaic
  id:   Int32 [bucket 1]
  flag: Int32 [bucket 0]
```
3. It's a single native binary, no JVM.

JingsongLi:
Please make sure your ci passed, you can run it in your local.

cli/src/main.rs
```diff
@@ -0,0 +1,481 @@
...
+            rows += batch.num_rows();
+            w.write_batch(&batch)?;
+        }
+        w.close()
+    })();
+    if let Err(e) = res {
+        let _ = std::fs::remove_file(&tmp);
+        return Err(e);
+    }
+    std::fs::rename(&tmp, out)?;
```
JingsongLi, line 504: Could we avoid replacing an existing output unless the user opts in? This rename silently replaces `out` on Unix if the path already exists; I reproduced `mosaic convert in.csv -o out.mosaic` overwriting a pre-existing file with exit code 0. Parquet CLI write commands expose `--overwrite`, so this default is surprising and potentially destructive. Please add an `--overwrite` flag and fail when `out` exists unless it is set.

cli/src/main.rs
```diff
@@ -0,0 +1,481 @@
...
+        json: bool,
+    },
+    /// Print per-column encoding and slot size for each row group.
+    Pages {
+        file: PathBuf,
+        #[arg(long)]
+        json: bool,
+    },
+    /// Print the first N rows as a table (alias: head).
+    #[command(visible_alias = "head")]
```
JingsongLi: Please don't model `head` as a pure alias of `cat`. In parquet-cli these are registered separately: `cat` defaults to no row limit, while `head` defaults to 10. With this alias, `mosaic cat file` truncates to 10 rows, which is a semantic mismatch for users coming from parquet-cli. I'd split the enum variants but keep a shared handler.

core/src/reader.rs
```diff
@@ -179,6 +179,30 @@ fn read_merged_ranges<I: InputFile + ?Sized>(
     Ok((merged, fetched))
 }
 
+/// Physical placement of one column within one row group.
+pub struct PageInfo {
```
JingsongLi, line 249: Can we avoid making this CLI/debug shape part of the stable core reader API? `PageInfo`/`BucketInfo` expose raw `u8` encoding codes and stringly bucket kinds, and the lower-level bucket readers now expose dictionary/encoding internals just for CLI rendering. Once public, downstream code can depend on these details. Consider a dedicated `inspect` module with typed enums/`#[non_exhaustive]` structs, or keep the raw accessors crate-private.

cli/src/main.rs
```diff
@@ -0,0 +1,481 @@
...
+        json: bool,
+    },
+    /// Print row-group / bucket / stats metadata.
+    Meta {
+        file: PathBuf,
+        #[arg(long)]
+        json: bool,
+    },
+    /// Print per-column encoding and slot size for each row group.
+    Pages {
```
JingsongLi, line 55: Please add `-c/--column(s)` filtering to `pages` and `column-size`, or make the divergence from parquet-cli explicit. Parquet CLI supports column selection for both commands, and this matters more for Mosaic wide tables because `pages` currently calls `page_infos` for every row group and may read/decompress all bucket data just to inspect one column.

jianguotian:
> Please make sure your ci passed, you can run it in your local.

I missed `cargo fmt` locally before. Now fixed it, and verified clippy + all rust/java/python tests pass locally.

cli/src/main.rs
jianguotian: fixed.

cli/src/main.rs
jianguotian: fixed.

core/src/reader.rs
jianguotian: fixed.

cli/src/main.rs
jianguotian: fixed.

cli/src/main.rs
```diff
@@ -0,0 +1,789 @@
...
+    }
+    // Column index of the filter target, for stats-based row-group skipping.
+    let pred_col = pred.as_ref().and_then(|p| {
+        reader
+            .schema()
+            .columns
+            .iter()
+            .position(|c| c.name == p.column)
+    });
+    let mut batches: Vec<RecordBatch> = Vec::new();
```
JingsongLi, line 565: Now that `cat` defaults to all rows, matching parquet-cli, the non-JSON path should not collect every row group into `batches`. Parquet CLI streams records as it reads them; this implementation makes `mosaic cat file` load the whole file before printing. Could we either stream text output per batch/row group, or reserve the global pretty table for bounded `head` / `-n` output?

cli/src/main.rs
```diff
@@ -0,0 +1,789 @@
...
+
+fn pages(file: &Path, columns: Option<String>, json: bool) -> std::io::Result<()> {
+    let reader = open(file)?;
+    let s = reader.schema();
+    let want = col_filter(&columns, s)?;
+    let nrg = reader.num_row_groups();
+    if json {
+        let mut row_groups = Vec::new();
+        for rg in 0..nrg {
+            let pgs = reader
```
JingsongLi, line 385: `-c/--columns` is currently only an output filter: `page_infos(rg)` has already inspected every column. In parquet-cli, column selection is applied before page iteration, so `pages -c one_col` avoids scanning unrelated columns. It would be better for the CLI to call a projected reader API, for example `page_infos_projected(rg, selected_indices)`, so the option's semantics and cost match the API surface.

core/src/reader.rs
```diff
@@ -447,6 +535,226 @@ impl<I: InputFile> MosaicReader<I> {
...
+                        } else {
+                            let slot = read_range(&self.input, foff, sizes[local])?;
+                            let ct = self.schema.columns[gi].data_type.clone();
+                            Self::parse_column_slot(&slot, &ct, meta.num_rows)?.encoding()
+                        };
+                        out.push(PageInfo {
+                            column_index: gi,
+                            bucket: b,
+                            encoding: Encoding::from_code(enc),
+                            slot_size: sizes[local],
```
JingsongLi: For paged buckets with ARRAY/LIST columns, `sizes[local]` is only the primary physical slot; child slots returned by `paged_dir()` are ignored here. `slot_sizes()` already attributes child bytes to the logical parent, so `pages` and `column-size` can disagree for nested columns. Please sum child slot sizes into the parent `PageInfo.slot_size` and add an assertion in the existing paged-array test.

cli/src/fmt.rs
```diff
@@ -0,0 +1,327 @@
...
+/// a single nanosecond count.
+pub fn render_json(v: &Value) -> String {
+    match v {
+        Value::Date(x) | Value::Time(x) => x.to_string(),
+        Value::TimestampMillis(x) | Value::TimestampMicros(x) => x.to_string(),
+        Value::TimestampNanos {
+            millis,
+            nanos_of_milli,
+        } => (*millis as i128 * 1_000_000 + *nanos_of_milli as i128).to_string(),
+        _ => render_value(v),
```
JingsongLi, line 61: This fallback routes string values through `render_value()`, which calls `safe()` and replaces control characters. JSON output should be lossless and machine-readable, leaving escaping to serde/Arrow; terminal sanitization should stay in text output only. Please keep raw strings for JSON, and apply the same principle to JSON column names such as `buckets --json`.

JingsongLi (commented):
Reviewed the latest revision with a focus on Parquet CLI alignment and API design.

JingsongLi (commented):
Follow-up review on the latest revision.

cli/src/main.rs
```diff
@@ -0,0 +1,823 @@
...
+    /// Print per-column encoding and slot size for each row group.
+    Pages {
+        file: PathBuf,
+        /// Comma-separated columns to show (default: all).
+        #[arg(short, long)]
+        columns: Option<String>,
+        #[arg(long)]
+        json: bool,
+    },
+    /// Print rows as a table (default: first 10; use --all to scan all).
```
JingsongLi: This reintroduces the Parquet CLI mismatch we fixed earlier. In parquet-cli, `cat` defaults to no row limit and `head` is the preview command with default 10 rows; adding `--all` makes `mosaic cat file` behave like `head`, so users coming from parquet-cli will silently see only a preview. Please keep `cat` unbounded by default and address memory usage by streaming/bounded table formatting rather than changing the command semantics. The docs/tests added in this commit should be updated accordingly too.

cli/src/main.rs
jianguotian: Fixed: unbounded text `cat` now streams per row group, while bounded output still buffers for table width.

cli/src/main.rs
jianguotian: Fixed: `pages -c` now uses `page_infos_projected`, so unrelated columns are not inspected.

core/src/reader.rs
jianguotian: Fixed: `PageInfo.slot_size` now includes ARRAY/LIST child slots, with a paged-array test assertion.

cli/src/fmt.rs
jianguotian: Fixed: JSON string values and bucket column names now stay lossless; only text output uses `safe()`.

cli/src/main.rs
jianguotian: Fixed: `cat` is unbounded by default again, and memory usage is handled by streaming output.

JingsongLi (commented):
Follow-up review on the latest revision.

cli/src/main.rs
```diff
@@ -0,0 +1,845 @@
...
+        #[arg(long)]
+        json: bool,
+    },
+    /// Print rows as a table (default: all rows; use -n to limit).
+    Cat {
+        file: PathBuf,
+        /// Limit to N rows.
+        #[arg(short = 'n', long)]
+        num: Option<usize>,
+        /// Print every row.
```
JingsongLi: Now that `cat` is unbounded by default again, this flag no longer adds a distinct mode; its only visible effect is that `mosaic cat -n 10 --all` silently ignores the limit. Parquet CLI keeps this surface smaller (`cat` = all rows, `head`/`-n` = bounded output). Could we either drop `--all`, or at least mark it as conflicting with `-n` so the API contract stays unambiguous?

JingsongLi:
+1

PR #69, opened 2026-07-02
[cli] Add schema-aware JSON and CSV conversion

jianguotian:
## Purpose

CSV is not self-describing, so an all-empty column is inferred as Arrow `Null` and cannot be written to Mosaic. The CLI conversion path also needs to preserve CSV parsing semantics, apply explicit schemas safely, and retain Avro logical and nested type semantics. It must also bound untrusted convert inputs so a single hostile record, an over-wide CSV, or an inflated schema file cannot exhaust memory before validation runs.

## Changes

- Keep `convert` for JSON input and add `convert-csv` for CSV input, including multi-file conversion and CSV dialect options.
- Convert all-empty inferred CSV columns to nullable `Utf8` while preserving their values as null; support `--require` and explicit Avro schemas when inference is insufficient.
- Make CSV escaping opt-in so RFC 4180 input preserves literal backslashes, and validate headers, empty inputs, missing required fields, and unsupported CSV schema types before publishing output.
- Accept comma-separated fields in every `-c`/`--column` occurrence and infer projected JSON schemas only from selected fields.
- Preserve UTC semantics for Avro `timestamp-*`, keep `local-timestamp-*` timezone-free, and recursively support Avro arrays and maps for JSON conversion.
- Reject explicit offsets in inferred local CSV timestamps instead of silently shifting the value and dropping its timezone; use `--schema` to select UTC instant semantics.
- Project only mapped columns for schema-driven wide CSV input, bound decoder and validation batches by payload size, and sanitize terminal-facing CSV schema errors.
- Bound attacker-controlled convert inputs before parsing: JSON root values are capped at 16 MiB, 100,000 structural units, and physical-line length; CSV is capped at 65,535 columns and 64 MiB of decoded field payload per record; and `--schema` files are capped at 4 MiB. Streaming readers enforce these limits before serde or Arrow materializes a record, on every inference and decode path.
- Validate value ranges before publishing output: Avro `time-millis` must fall in `[0, 86400000)` on both the JSON and CSV paths, and integer-shaped tokens in inferred `Float64` columns must round-trip exactly instead of being silently rounded.
- Update the CLI README, website documentation, and end-to-end coverage.

## Compatibility

- `convert` accepts JSON inputs only; scripts built from unreleased `main` that used `mosaic convert data.csv` should use `mosaic convert-csv data.csv` instead.
- Dropped the never-shipped `--out` and `--columns` aliases (added only on unreleased `main`); use `-o`/`--output` and `-c`/`--column`.
- The Mosaic CLI was added after the v0.2.0 release, so this command split has not shipped in a published release.
- Inferred CSV timestamps are local timestamps. Values with explicit timezone offsets fail fast; provide an Avro `timestamp-*` schema when the column represents UTC instants.

## Tests

- `cargo deny check licenses`
- `cargo build --all-targets`
- `cargo test -p paimon-mosaic-cli` (64 unit + 93 e2e tests)
- `cargo test --workspace -- --skip test_read_java_written_file --skip test_read_python_written_file`
- `cargo llvm-cov --summary-only -- --skip test_read_java_written_file --skip test_read_python_written_file`
- The control-character regression tests failed against the pre-fix error formatting and pass after sanitizing the remaining field names.
- A synthetic merge with the current `main` on August 21, 2026 passed formatting, workspace Clippy, CLI tests, license validation, and ASF YAML validation.
- Manual CLI smoke test for all-empty CSV columns, literal backslashes, `schema`/`cat`/`count`, and `-c id,kind` JSON projection.


cli/src/main.rs
```diff
@@ -518,36 +1343,1623 @@ fn convert(
...
+        self.saw_int64 && self.saw_float64
+    }
+}
+
+fn observe_csv_inferred_types(
+    inferred_types: &mut std::collections::HashMap<String, ObservedCsvTypes>,
+    schema: &Schema,
+) {
+    for field in schema.fields() {
+        if matches!(field.data_type(), DataType::Int64 | DataType::Float64) {
```
JingsongLi: [P1] Preserve integer tokens inside Float64-inferred shards

This records only the final inferred type for each shard. If a shard contains both 9007199254740993 and 1.5, Arrow already reports Float64, so saw_int64 remains false and mixed_csv_float_fields does not route the column through the raw-text exactness check. With another Float64 shard, conversion succeeds but stores 9007199254740992, silently changing a valid Int64 token. Please retain per-token integer observations during inference, or read every inferred Float64 column as Utf8 and apply the existing exact-integral-token check. A regression test should cover multiple inputs where no shard infers as pure Int64.

cli/src/main.rs
```diff
@@ -429,48 +530,777 @@ fn count(file: &Path, json: bool) -> std::io::Result<()> {
...
+            invalid_schema(format!(
+                "invalid UUID '{}' for JSON field '{}' at record {record}",
+                fmt::safe(&value),
+                fmt::safe(path)
+            ))
+        })?;
+    }
+    let data_type = field.data_type();
+    match data_type {
+        DataType::Int32 | DataType::Date32 | DataType::Time32(_) if !raw.get().starts_with('"') => {
```
JingsongLi: [P1] Enforce the Avro time-millis day range

Time32(Millisecond) is validated only against i32 bounds here, but Avro time-millis is milliseconds after midnight, so valid values are 0 <= value < 86400000. Inputs -1 and 86400000 currently convert successfully and persist invalid Time32 values; cat --json later emits a temporal cast error. Please validate this type separately and add boundary coverage for -1, 86399999, and 86400000. See the Avro specification: https://avro.apache.org/docs/1.11.2/specification/#time-millisecond-precision

cli/src/main.rs
```diff
@@ -429,48 +530,815 @@ fn count(file: &Path, json: bool) -> std::io::Result<()> {
...
+    let normalize_decimals = schema_has_decimal(schema);
+    let mut decoder = build_decoder()?;
+    let mut batch_bytes = 0_usize;
+    let records = serde_json::Deserializer::from_reader(reader).into_iter::<Box<RawValue>>();
+
+    for (index, raw) in records.enumerate() {
+        let record = index + 1;
+        let raw = raw.map_err(|e| invalid_schema(format!("invalid JSON record {record}: {e}")))?;
+        let raw_bytes = raw.get().as_bytes();
+        if raw_bytes.len() > MAX_JSON_RECORD_BYTES {
```
JingsongLi: [P2] Enforce the record limit before allocation on every JSON path

The limit is checked only after serde_json has materialized the complete Box<RawValue>, so a record larger than available memory can exhaust the process before reaching this branch. In addition, inferred schemas and explicit schemas without special-value validation take the Arrow reader path at lines 554-568 and never consult MAX_JSON_RECORD_BYTES. Please put the limit in a bounded record reader shared by inference and decoding so it fires before either parser allocates the complete record.

cli/src/main.rs
```diff
@@ -429,48 +530,815 @@ fn count(file: &Path, json: bool) -> std::io::Result<()> {
...
+    let explicit_schema = schema.map(load_convert_schema).transpose()?;
+    let has_explicit_schema = explicit_schema.is_some();
+    let schema = match explicit_schema {
+        Some(schema) => schema,
+        None => {
+            let mut inferred: Option<Schema> = None;
+            for input in inputs {
+                let (reader, line_offset) = open_csv(input, options.skip_lines)?;
+                let (schema, rows) = format
+                    .infer_schema(reader, None)
```
JingsongLi: [P2] Reject oversized CSV width before schema inference

For inferred CSVs, format.infer_schema runs before open_csv_input applies MAX_CSV_COLUMNS. A wide input can therefore make Arrow allocate names, types, and Fields for every column before the rejection. The positional path also sets layout.columns to zero for --no-header, so it bypasses the limit altogether. Please validate the first header or data record width before inference and carry the observed width through the positional layout.

cli/src/main.rs
jianguotian: Fixed. Every inferred `Float64` column is now read as raw text and each integer token must round-trip exactly, so `9007199254740993` is rejected instead of rounded, even when no shard infers as pure `Int64`. Regression test: `convert_csv_inferred_float_rejects_lossy_bare_integer_literals` (both input orders).

cli/src/main.rs
jianguotian: Fixed. `Time32(millis)` is now validated against the Avro day range `0..=86_399_999` (not just `i32`) on both the JSON and CSV paths. Boundaries `-1 / 86399999 / 86400000` are covered by `convert_json_enforces_avro_time_millis_day_range` and `convert_csv_enforces_avro_time_millis_day_range`.

cli/src/main.rs
jianguotian: Fixed. The limit now lives in a streaming `JsonRecordLimitReader` shared by both inference and decoding, so it fires during `read()` before serde or Arrow materializes the record, on all JSON paths (inferred, projected, plain Arrow reader, validated). Test: `json_record_limit_stops_at_limit_plus_one_without_reading_the_tail`.

cli/src/main.rs
jianguotian: Fixed. A `CsvRecordLimitScanner` enforces the column limit per record in a pass that runs before inference, so Arrow never allocates fields for an over-wide header/record; it counts fields per record regardless of `--header`/`--no-header`, closing the positional bypass. Test: `open_csv_input_rejects_too_many_columns`.

PR #70, opened 2026-07-06
fix(core): validate writer batch schema

QuakeWang:
## Problem

`MosaicWriter` builds `batch_col_map` from the schema supplied at construction time. A later `RecordBatch` can have the same number of columns and compatible Arrow types but different field names or order. Reusing the cached map in that case can write values under the wrong logical columns without an error.

## Changes

- Validate each mapped top-level field name before writing a batch.
- Reject reordered or renamed fields with `InvalidInput`.
- Report the actual mapped batch column index in the validation error.
- Add regression coverage for reordered schemas and mapped-index diagnostics.

## Scope

This PR adds the missing field-name and field-order guard. Arrow array type validation and actual non-null enforcement remain on the existing writer validation path. Nullable Arrow field metadata is intentionally accepted when the arrays satisfy the writer schema.

## Validation

- `test_write_batch_rejects_reordered_schema`
- `test_write_batch_name_mismatch_reports_mapped_batch_index`
- [GitHub Actions CI](https://github.com/apache/paimon-mosaic/actions/runs/32353562910)

### DIFF
--- a/core/src/writer.rs
+++ b/core/src/writer.rs
@@ -18,7 +18,7 @@
 use std::io;
 
 use arrow_array::*;
-use arrow_schema::Schema;
+use arrow_schema::{DataType, Field, Schema};
 
 use crate::bucket_writer::{BucketWriter, PagedBucketOutput};
 use crate::schema::MosaicSchema;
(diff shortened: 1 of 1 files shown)


jianguotian:
I found and fixed two regressions in the current head:

- keep field-name/order validation, but leave type/nullability enforcement on the existing writer validation path so current error/test contracts remain intact;
- report the actual mapped batch column index in the name-mismatch error.

Because I cannot push to `QuakeWang:mosaic-schema-guard`, the fix is available as `fbaf0a0b5df19f5d7141e4b78ccfbc17601c1215` and helper PR https://github.com/QuakeWang/paimon-mosaic/pull/1.

Validation:
- local full Rust/Python/C++/Java suites, fmt and Clippy passed;
- exact-head GitHub CI against current Apache main `b613dd25eba45afbc860171e109df19495ba0e3c`: all 5 jobs passed: https://github.com/jianguotian/paimon-mosaic/actions/runs/32209856324.

Please merge the helper PR (or cherry-pick the commit) to update this PR.

QuakeWang:
The follow-up changes from QuakeWang/paimon-mosaic#1 have been merged into `75eeeffe`.

The final scope is limited to validating mapped top-level field names before writing, which prevents same-typed reordered batches from being routed to the wrong logical columns. The existing writer validation continues to handle Arrow array types and actual null values.

All five CI jobs are green: https://github.com/apache/paimon-mosaic/actions/runs/32353562910.

The PR is ready for another review.

jianguotian:
For the 0.3.0 release, I propose treating this as a correctness blocker: a reordered batch with compatible types can otherwise be written under the wrong logical columns without an error. The current head is mergeable and all five CI jobs are green.

PR #71, opened 2026-07-08
Move Java staging deployment to a hardened local release flow

jianguotian:
## Purpose

Move Java RC signing and Nexus staging deployment from GitHub Actions to the Release Manager's machine, while binding the local deployment to the exact RC tag and successful Release workflow run.

## Changes

- Keep Java signing keys and Nexus credentials out of GitHub Actions; CI packages and uploads the Java/native artifacts for local staging.
- Harden argument parsing so a missing option value cannot consume `--dry-run` and trigger a real deployment.
- Require the release version, RC number, tag, workflow ref, workflow commit, and local checkout to match exactly.
- Reject tracked, untracked, and ignored Java package inputs that could enter release JARs; normalize relative Maven settings paths.
- Support macOS Bash 3.2 and run the regression suite on both Ubuntu and macOS.
- Validate the exact main, sources, javadoc, and native-containing JARs during the same `mvn clean deploy` lifecycle that uploads them.
- Update the official release guide for the local Java staging procedure.
- Fix the current Rust Clippy `byte_char_slices` failure so the PR can become green.

## Tests

- `bash tools/tests/deploy_java_staging_test.sh` (7 safety regressions)
- The same suite with GNU Bash 3.2
- `mvn clean verify -Prelease -Dgpg.skip=true -DskipTests -DstagingValidationScript=...` on JDK 8
- `cargo +1.97.1 fmt --all -- --check`
- `cargo +1.97.1 clippy --all-targets --workspace -- -D warnings`
- Parsed all workflow YAML and the release HTML
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,25 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  release-script-test:
+    name: Release script (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Check shell syntax
+        run: |
+          /bin/bash -n tools/deploy_java_staging.sh
+          /bin/bash -n tools/validate_java_staging_artifacts.sh
+          /bin/bash -n tools/tests/deploy_java_staging_test.sh
+
+      - name: Test Java staging deploy safety
+        run: /bin/bash tools/tests/deploy_java_staging_test.sh
+
   check:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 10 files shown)


jianguotian:
Implemented the release-safety fixes in `0091ca3`:

- fail closed on missing option values (including the `--dry-run` regression)
- reject ignored package inputs
- bind the workflow run ref/SHA to the exact RC tag and version/RC tuple
- support Bash 3.2 and normalize Maven settings paths
- validate the exact JARs in the same Maven lifecycle that deploys them
- update the release guide and add Ubuntu/macOS regression coverage
- fix the current Clippy failure

Local validation passed (7 shell regressions on Bash 4.4 and Bash 3.2, JDK 8 Maven artifact validation, Rust fmt/Clippy, workflow YAML/HTML parsing, and `git diff --check`).

The new CI run is waiting for an upstream maintainer to approve the fork workflow: https://github.com/apache/paimon-mosaic/actions/runs/29577101080

PR #73, opened 2026-08-06
[java] Support Arrow batch writes across allocator roots

jianguotian:
## Purpose

`MosaicWriter.write(VectorSchemaRoot)` exports incoming Arrow batches through the Arrow C Data interface. Arrow 15's `Data.exportVectorSchemaRoot` reloads the fields into a temporary root struct, which cannot reload buffers owned by an independent allocator root into the writer allocator.

Support those batches without moving or reassociating their buffers: keep the existing same-root path, export cross-root fields directly, and retain temporary Arrow C Data metadata under the writer allocator. Make failed cross-root exports rollback-safe so an allocator OOM or partial child export does not retain input buffers.

## Changes

- Recursively validate that all top-level and nested input vectors share one input allocator root.
- Keep same-root batches on the existing `Data.exportVectorSchemaRoot` path.
- Export cross-root schemas and field arrays directly while charging temporary C Data metadata to the writer allocator.
- Release all successfully exported children and temporary metadata when a later cross-root export step fails.
- Document the synchronous batch lifetime, allocator-root requirement, and writer-owned metadata accounting.
- Keep the native writer API, Rust codec, and on-disk format unchanged.

## Tests

- JDK 8: 55 tests passed, 0 failed, 0 errors, 0 skipped.
- JDK 11: 55 tests passed, 0 failed, 0 errors, 0 skipped.
- Covered independent roots, same-root accounting, sequential roots, nested mixed-root validation, writer allocator OOM retry, and partial child-export rollback.
- `cargo fmt --all -- --check` passed.
- `git diff --check` passed.

### DIFF
--- a/java/src/main/java/org/apache/paimon/mosaic/MosaicWriter.java
+++ b/java/src/main/java/org/apache/paimon/mosaic/MosaicWriter.java
@@ -105,6 +105,8 @@ public class MosaicWriter implements AutoCloseable {
                 releaseExported(arrowArray);
                 releaseExported(arrowSchema);
             }
+        } catch (Throwable failure) {
+            throw propagateNativeFailure("write batch failed", failure);
         }
     }
 
@@ -233,6 +235,17 @@ public class MosaicWriter implements AutoCloseable {
         }
     }
 
+    private static RuntimeException propagateNativeFailure(
+            String message, Throwable failure) {
+        if (failure instanceof RuntimeException) {
+            return (RuntimeException) failure;
+        }
+        if (failure instanceof Error) {
+            throw (Error) failure;
+        }
+        return new RuntimeException(message, failure);
+    }
+
     public long estimatedFileSize() {
         return NativeLib.nativeWriterEstimatedSize(handle);
     }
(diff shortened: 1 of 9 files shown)


java/src/main/java/org/apache/paimon/mosaic/MosaicWriter.java
```diff
@@ -69,19 +73,153 @@ public MosaicWriter(OutputStream outputStream, Schema arrowSchema, WriterOptions
...
+            snapshot.children =
+                    privateData.childPointers == null
+                            ? 0
+                            : privateData.childPointers.memoryAddress();
+            snapshot.dictionary = 0;
+            snapshot.release = 0;
+            arrowArray.save(snapshot);
+            JniWrapper.get().exportArray(arrowArray.memoryAddress(), privateData);
+        } catch (RuntimeException | Error failure) {
+            privateData.abort(failure);
```
JingsongLi: [P1] Release the registered root callback before aborting child metadata

`JniWrapper.exportArray` can return with an `OutOfMemoryError` after Arrow 15 has already installed the root `ArrowArray` release callback: its JNI code does not check the result of `NewGlobalRef(privateData)` before assigning `private_data` and `release`. In that state, `privateData.abort()` frees `childPointers` and the child structs, but the outer `finally` then invokes the still-registered root callback, whose first step is to walk `root.children`. That dereferences freed native memory and can crash or corrupt the JVM under memory pressure.

Please branch on root ownership here: when `arrowArray.snapshot().release != 0`, release the root first while its child pointer table is still valid, then call the idempotent `privateData.close()` as a fallback. Use the current manual `abort()` path only while the root release callback is still zero. A fault-injection test that throws after root callback registration would cover this window.

JingsongLi:
+1

PR #75, opened 2026-08-19
perf(core): speed up reading nullable fixed-width CONST columns

jianguotian:
## Summary

A fixed-width `CONST` column stores one value for all non-null rows. When such a column also contained nulls, the reader first decoded a compact value buffer and then copied those values into their final row positions.

This PR builds the final Arrow value buffer directly, avoiding the extra allocation and copy.

## Changes

- Build row-aligned Arrow values directly for fixed-width `CONST` columns with nulls.
- Below 1/8 non-null density, write only valid positions.
- At ordinary densities, bulk-fill the whole buffer only when every physical OS page covered by the actual output allocation contains a valid row; otherwise bulk-fill only contiguous non-null row runs.
- Never use logical chunk boundaries as write ranges, because `Vec` allocations are not guaranteed to align with physical pages.
- Cover nullable `INT16(0)` through the generic `Int16` path; there is no `INT16`- or zero-specific implementation.
- Preserve zero values, floating-point negative zero, Boolean values, timestamps, and trailing bitmap semantics.
- Ignore hidden payloads at null positions when converting nullable row-aligned nanosecond timestamps, while keeping PLAIN, DICT, and non-null timestamp conversion on the unchecked path.
- Cover sparse nullable round trips for every fixed-width variant across monolithic and paged readers.
- Keep variable-width `Utf8`, `Binary`, and large-decimal constants on the existing compact path.
- No file-format or public API changes.

### Wide-table reader benchmark

An earlier optimization-head A/B measurement, before the final RSS guard, used 12 real-world wide-table Mosaic files:

| Test input | Value |
|---|---:|
| Files | 12 |
| Columns per file | 8,691 |
| Total rows | 115,631 |
| Runs | 7 per file, alternating base/head order |

| Metric across 12 files | Base | Optimized | Change |
|---|---:|---:|---:|
| Read all columns | 1,636.63 ms | 989.78 ms | **-39.5%** |
| Open reader + read all columns | 1,702.51 ms | 1,055.68 ms | **-38.0%** |

- All 12 files improved; per-file column-read reduction was **35.6% to 42.3%**.
- Base and optimized runs produced identical row counts, column counts, and null counts.

### TimestampNanos hot path

A pinned-CPU release probe converted 16,777,216 non-null nanosecond timestamps over 7 runs:

| Version | Task clock | Instructions |
|---|---:|---:|
| Base | 94.70 ms | 623.04M |
| Optimized | 94.66 ms | 623.03M |

The null-aware conversion is selected outside the row loop and only for nullable row-aligned values, so ordinary PLAIN, DICT, and non-null reads retain the original hot path.

### Page-amplification regression

A 3-run exact-loop Rust probe on an aarch64/glibc host used 32M `Int64` rows at exactly 1/8 non-null density, with alternating touched/null-only logical chunks:

| Materialization path | Time | Peak RSS |
|---|---:|---:|
| Original compact + scatter | 62-63 ms | 169,520-169,528 KiB |
| Page-amplified touched chunks | 56-58 ms | 267,728-267,736 KiB |
| Final non-null row runs | **33-34 ms** | **136,792 KiB** |

These are local Core/materialization measurements, not end-to-end capacity or CU results.

## Validation

- `cargo deny check licenses`
- `cargo +1.86 check -p paimon-mosaic-core --all-targets`
- Workspace tests: **397 passed**; the two readback cases requiring externally generated Java/Python files were skipped locally
- Sparse row-alignment mutation checks: removing either `Int8` or `TimestampNanos` from the direct path is rejected by the round-trip test
- Nullable nanosecond timestamp allocator-layout regression probe: **20,000 reads passed**
- Pinned-CPU nanosecond timestamp probe: non-null task clock and instruction count match base
- Remote CI status is tracked by the current PR checks

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -30,3 +30,6 @@ zstd = "0.13"
 arrow-schema = "58"
 arrow-array = "58"
 arrow-buffer = "58"
+
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
--- a/core/src/bucket_reader.rs
+++ b/core/src/bucket_reader.rs
@@ -16,6 +16,7 @@
 // under the License.
 
 use std::io;
+use std::mem::size_of;
 use std::sync::Arc;
 
 use arrow_array::*;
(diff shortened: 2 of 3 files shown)


JingsongLi (approved):
+1

PR #76, opened 2026-08-20
feat(core): add encoded column access API without Arrow arrays

jianguotian:
## Summary

Add public Core types and a `RowGroupReader` API for accessing non-nested Mosaic columns in their existing encoding without first creating Arrow arrays.

The existing `read_columns()` path, projection behavior, and file format remain unchanged.

### Public types

- Add `EncodedColumn`, a view of a column's type, encoding, row count, null bitmap, and encoded data.
- Add `EncodedColumnValues`, an iterator over values in row order.
- Add `EncodedValueRef`, which represents null, numeric, string, binary, decimal, date, time, and timestamp values without owning variable-length data.

### Reader API

- Add `RowGroupReader::visit_encoded_columns`.
- Visit projected columns in output order.
- Pass the column name, data type, nullability, and `EncodedColumn` to the consumer.
- Preserve the existing projection order and repeated-projection behavior.

### Encoding support

- `ALL_NULL`: identify the whole column as null without reading individual values.
- `CONST`: expose the single stored value and null bitmap.
- `DICT`: read dictionary indexes and return the corresponding values.
- `PLAIN`: parse values in row order.
- Support both monolithic and paged buckets.
- Reject projected List and Map columns before invoking the consumer.
- Return errors for invalid encodings, truncated values, and invalid dictionary indexes.

## Performance

Tested with a customer sample containing:

- 9,837 rows and 8,690 columns;
- 88.72% `CONST` or `ALL_NULL` columns;
- 100 monolithic buckets using Zstd.

Release build, three independent runs with 11 iterations each, reporting the median:

| Path | Processing time | Allocated memory |
|---|---:|---:|
| Existing `read_columns()` path | 155.97 ms | 261.1 MiB |
| New `visit_encoded_columns()` path | 81.00 ms | 8.5 KiB |

The new API was used to handle `ALL_NULL` and `CONST` columns directly and iterate the values of `DICT` and `PLAIN` columns. For this sample, column processing time decreased by 48.1%.

This benchmark measures column processing after the row group has been loaded. It does not include file I/O, bucket decompression, JSON generation, or final output compression.

## Tests

- Compare values returned by the new API with `read_columns()`.
- Cover all supported non-nested column types and encodings.
- Cover monolithic and paged buckets.
- Verify projection order and repeated projections.
- Verify errors for invalid data and unsupported nested columns.

## Validation

- `cargo +1.97.1 fmt --all -- --check`
- `cargo +1.97.1 test -p paimon-mosaic-core` — 371 passed
- `cargo +1.97.1 clippy --all-targets --workspace -- -D warnings`
- `cargo +1.86 check -p paimon-mosaic-core --all-targets`
- `cargo deny check licenses`
- `git diff --check`

### DIFF
--- a/core/src/bucket_reader.rs
+++ b/core/src/bucket_reader.rs
@@ -23,10 +23,334 @@ use arrow_array::*;
 use arrow_buffer::{BooleanBuffer, Buffer, NullBuffer, OffsetBuffer, ScalarBuffer};
 use arrow_schema::{DataType, Field, TimeUnit};
 
+use crate::reader::Encoding;
 use crate::spec::*;
 use crate::types;
 use crate::values::Value;
 use crate::varint;
+
+/// Borrowed view of one encoded scalar column.
+///
+/// The view borrows buffers owned by a [`crate::reader::RowGroupReader`] and must be consumed
+/// synchronously during [`crate::reader::RowGroupReader::visit_encoded_columns`]. It exposes the
+/// physical encoding without first materializing an Arrow array.
+#[derive(Clone, Copy)]
+pub struct EncodedColumn<'a> {
+    data_type: &'a DataType,
+    encoding: u8,
+    has_nulls: bool,
+    null_bitmap: &'a [u8],
+    const_value: &'a Value,
+    dict_values: &'a [Value],
+    dict_bit_width: usize,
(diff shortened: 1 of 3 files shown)


JingsongLi:
+1

PR #77, opened 2026-08-20
fix(jni): propagate InputFile exceptions from native reader threads

jianguotian:
## Summary

Propagate the original Java exception thrown by `InputFile.readFully(...)` back to the caller when Mosaic reads through native worker threads, instead of replacing it with a generic `RuntimeException`.

## Changes

- Capture and clear worker-thread JNI exceptions before detaching, then rethrow the same throwable on the calling Java thread.
- Handle failures during reader open and row-group reads without leaking JNI global references.
- Declare `IOException` on `MosaicReader.open(...)` and `readRowGroup(...)` and document the contract.

The Mosaic file format is unchanged. This is a source-incompatible Java API change for 0.3: callers must catch or declare `IOException`, and affected method references or lambdas must be updated.

## Validation

Validated at `e4833d12b8d574911c021335e4ceeacd17c8de85` on current-main synthetic merge `525c717241da9de4bd2e688cd15951cb24893332`:

- Rust format and Clippy passed.
- Rust workspace release tests: 405 passed.
- Java 11 tests: 62 passed.
- JNI release build and Javadoc generation passed.

### DIFF
--- a/docs/java-api.html
+++ b/docs/java-api.html
@@ -260,7 +260,13 @@ w.endMap();</code></pre>
 <pre><code><span class="ty">BufferAllocator</span> allocator = <span class="kw">new</span> <span class="ty">RootAllocator</span>();
 <span class="ty">InputFile</span> inputFile = ...;
 <span class="kw">long</span> fileLength = ...;
+<span class="cmt">// The enclosing method declares throws IOException.</span>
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #79, opened 2026-08-21
feat(release): verify source and binary artifacts

jianguotian:
## Summary

- verify signed release tags, cross-language versions, reproducible source archives, Java JARs, Python wheels, and native JNI/FFI payloads;
- fail closed on malformed, oversized, redirected, incomplete, or unexpected release payloads;
- share ZIP admission rules across JAR and wheel verification and bound native parser work;
- gate Rust, Java, and Python release jobs behind one reusable tag/version preflight;
- make direct Rust/Java `workflow_dispatch` paths run the same preflight, so publication cannot bypass it;
- run the full wheel verifier before PyPI publication and reject every non-wheel artifact in `dist/`.

## Review closure

This head incorporates the verifier review fixes tracked in jianguotian/paimon-mosaic#22, including:

- callable Mach-O export validation and bounded ELF/PE/Mach-O structural work;
- exact Java class/source and wheel module/dependency/payload contracts;
- scrubbed Git repository environment for source archives;
- bounded wheel `RECORD` processing and bounded diagnostics;
- workflow-routing tests that kill validator, dependency, direct-dispatch, and publish-gate bypass mutations.

## Validation

- `python3 -m pytest -q tools/tests`: **345 passed, 48 subtests passed**
- **19/19** targeted workflow and verifier mutations killed
- `actionlint 1.7.12 .github/workflows/*.yml`: passed
- ASF header check and `python3 -m compileall -q tools`: passed
- current Apache `main` (`c4d8306`) synthetic merge: conflict-free
- synthetic merge: `cargo fmt --all -- --check` and `cargo check --locked --workspace --all-targets` passed
- fork CI run `32638325773`: **5/5 jobs passed**

## Exact heads

- Apache PR head: `05755fcd52c3d36217acba55e0668b3b0e10ddf9`
- fork validation PR head: `5c749bba8b1f2e493d89a22a842f5fc925bf7cfe`
- both heads have the exact same Git tree: `61eafd740ed6f95b642dd698d284cd3be8a6eeb9`

## Current status

The code and fork CI are green. Apache CI run `32638540033` is `action_required` with zero jobs until a maintainer approves execution; maintainer review/approval is also still pending.


tools/verify_release_versions.py
```diff
@@ -0,0 +1,687 @@
...
+    compatibility = {}
+    for dependency in path_dependencies(root, packages):
+        if dependency.requirement is None:
+            continue
+        key = (dependency.requirement, dependency.target.version)
+        if key not in compatibility:
+            compatibility[key] = cargo_requirement_accepts(*key)
+        if compatibility[key]:
+            continue
+        manifest = dependency.manifest.relative_to(root)
```
JingsongLi, line 256: [P2] Canonicalize root before relative path comparisons

Cargo returns canonical manifest paths, while root is used verbatim here. On macOS a temporary root commonly appears as /var/..., but Cargo reports /private/var/..., so relative_to(root) raises ValueError instead of returning version diagnostics. The same issue exists in update_cargo_versions. Please resolve root once at each public entry point before comparing Cargo paths.

tools/tests/test_validate_release_tag.py
```diff
@@ -0,0 +1,276 @@
...
+    assert validator.main() == 1
+    captured = capsys.readouterr()
+    assert "release tag validation failed" in captured.err
+
+
+def test_invalid_extra_rc_does_not_hide_a_valid_matching_rc(tmp_path, signing_keys):
+    (home, fingerprint, keys), _ = signing_keys
+    repo = repository(tmp_path)
+    sign_tag(repo, "v1.3.0-rc1", home, fingerprint)
+    run(["git", "tag", "v1.3.0-rc2"], cwd=repo)
```
JingsongLi: [P2] Make the lightweight-tag fixture ignore user signing config

This bare git tag inherits global tag.gpgSign. For release managers with tag.gpgSign=true, Git attempts an annotated signed tag and the noninteractive test fails with "Terminal is dumb, but EDITOR unset"; the same applies to the fixture at line 245. Please explicitly disable signing for intentionally lightweight tags, for example with git -c tag.gpgSign=false tag ...

tools/tests/test_validate_release_tag.py
```diff
@@ -0,0 +1,276 @@
...
+        cwd=tmp_path,
+        env=env,
+    )
+    keys.write_text(exported + "\n", encoding="utf-8")
+    return home, fingerprint, keys
+
+
+@pytest.fixture(scope="module")
+def signing_keys(tmp_path_factory):
+    root = tmp_path_factory.mktemp("release-signing-keys")
```
JingsongLi: [P2] Use a short temporary base for GNUPGHOME

On macOS the default pytest temporary directory is already long. Nesting both the signing-key fixture and GNUPGHOME below it exceeds the gpg-agent Unix socket path limit, causing seven tests to fail during setup with "File name too long". Please allocate these GPG homes under a deliberately short canonical temporary directory and clean it afterward.

tools/update_branch_version.sh
```diff
@@ -17,55 +17,110 @@
...
-#change version in all pom files (match both exact and -SNAPSHOT suffix)
-find . -name 'pom.xml' -type f -exec perl -pi -e 's#<version>'$OLD_VERSION'(-SNAPSHOT)?</version>#<version>'$NEW_VERSION'</version>#' {} \;
-
-#change version in Cargo.toml files
-find . -name 'Cargo.toml' -not -path '*/target/*' -type f -exec perl -pi -e 's#^version = "'$OLD_VERSION_CLEAN'"#version = "'$NEW_VERSION_CLEAN'"#' {} \;
+# Change the project version in all pom files. This is structural rather than
+# textual: a dependency or plugin element carrying the same version string must
+# not be rewritten along with it.
+find . -name 'pom.xml' -not -path '*/target/*' -type f \
+  -exec python3 tools/bump_pom_version.py "$OLD_VERSION" "$NEW_VERSION" {} +
```
JingsongLi, line 65: [P1] Preserve the documented clean-to-SNAPSHOT version bump

The release guide invokes this script with OLD_VERSION=0.3.0 and NEW_VERSION=0.4.0-SNAPSHOT and explicitly says that both 0.3.0 and 0.3.0-SNAPSHOT are matched. The current java/pom.xml contains 0.3.0-SNAPSHOT, but this now passes the raw clean OLD_VERSION to the exact structural matcher, so the documented command fails with "no project or parent version element holds 0.3.0". Please preserve structural scoping while accepting both the clean old version and its SNAPSHOT form, and add a regression test using the documented invocation.

tools/native_binary.py
```diff
@@ -0,0 +1,1724 @@
...
+                        terminal_end,
+                        "Mach-O export trie resolver address",
+                    )
+            if cursor != terminal_end:
+                raise ValueError(
+                    "Mach-O export trie terminal has trailing payload data"
+                )
+            name = ascii_symbol(prefix)
+            if name:
+                exported_symbols.add(name)
```
JingsongLi: [P1] Count only callable Mach-O exports

Every terminal name is added to exported_symbols regardless of its export kind or address. As a result, required JNI names represented only by absolute, thread-local, or re-export terminals satisfy verify_native_target even though they do not define callable code in this dylib. A synthetic dylib whose 24 required JNI names were absolute exports at address 1 passed verification. Please reject non-callable kinds and validate regular or stub addresses against an executable segment before counting them.

tools/verify_python_wheels.py
```diff
@@ -0,0 +1,547 @@
...
+        raise ValueError(
+            f"wheel declares more than {MAX_ARCHIVE_ENTRIES} entries: {len(entries)}"
+        )
+    for info in entries:
+        raw_name = info.orig_filename
+        if not raw_name or "\x00" in raw_name or raw_name != info.filename:
+            raise ValueError(f"invalid wheel entry path: {raw_name!r}")
+        if "\\" in raw_name:
+            raise ValueError(f"wheel entry uses a backslash: {raw_name!r}")
+        if raw_name.startswith("/") or WINDOWS_ABSOLUTE_PATH.match(raw_name):
```
JingsongLi: [P1] Reject Windows drive-relative wheel paths

The current pattern rejects C:/... but not drive-relative paths such as C:../site-packages_evil/payload.py. That name has no standalone .. component, so it passes this validation and can be listed consistently in RECORD; under Windows path semantics, joining it below site-packages normalizes one directory upward. Please reject every drive-qualified path, for example with ^[A-Za-z]:, or validate names with PureWindowsPath drive semantics.

tools/tests/test_validate_release_tag.py
```diff
@@ -0,0 +1,383 @@
...
+    revocation = home.parent / f"{fingerprint}.revoke.asc"
+    revocation.write_text(armored + "\n", encoding="utf-8")
+    env = os.environ.copy()
+    env["GNUPGHOME"] = str(home)
+    run(["gpg", "--batch", "--yes", "--import", str(revocation)], cwd=home, env=env)
+
+
+def test_tag_signed_by_a_revoked_key_is_rejected(tmp_path):
+    root = Path(tmp_path).resolve()
+    home, fingerprint, keys = generate_key(root, "Compromised Release")
```
JingsongLi: [P2] Keep the revoked-key GNUPGHOME under the short temporary root

The module signing-key fixture now uses a short path, but this new test calls generate_key with the regular pytest tmp_path and recreates the same Unix-socket overflow. On the default macOS temporary directory, key generation fails here with "gpg-agent: File name too long" before the revocation assertion runs; it still fails with TMPDIR=/private/tmp because of the additional pytest nesting. Please create this independent key under a TemporaryDirectory rooted at short_temporary_root().

tools/tests/test_verify_release_versions.py
```diff
@@ -0,0 +1,487 @@
...
+            "^0.3.0",
+            ">=0.3.0, <0.4.0",
+        )
+
+        updated = verify_release_versions.update_cargo_versions(
+            self.root, "0.3.0", "0.4.0"
+        )
+
+        self.assertEqual(
+            {path.relative_to(self.root).as_posix() for path in updated},
```
JingsongLi, line 152: [P2] Relativize updater results against the canonical fixture root

update_cargo_versions now intentionally resolves its root and returns canonical /private/var/... paths on macOS, while self.root remains the unresolved /var/... spelling. This relative_to call therefore raises ValueError under the default macOS TMPDIR instead of checking the returned manifests. Please resolve self.root in setup or relativize against self.root.resolve().

tools/verify_java_jars.py
```diff
@@ -0,0 +1,386 @@
...
+
+def native_archive_entries(
+    archive: ZipFile, entries: dict[str, ZipInfo]
+) -> set[str]:
+    native_entries = set()
+    for name, info in entries.items():
+        if info.is_dir():
+            continue
+        with archive.open(info) as source:
+            magic = native_binary_magic(source, info.file_size, name)
```
JingsongLi, line 308: [P2] Read every JAR member through EOF to validate its CRC

For ordinary resources, native_binary_magic reads only the small header and this context closes the member before EOF, so zipfile never verifies its CRC. Corrupting a stored classifier resource after its first 128 KiB made ZipFile.read raise BadZipFile, while verify_classifier still reported success. Please stream each member through EOF once, or run an equivalent complete CRC pass, before accepting the artifact.

tools/native_binary.py
```diff
@@ -0,0 +1,1724 @@
...
+    chains: tuple[int, ...]
+
+    @property
+    def symbol_count(self) -> int:
+        return len(self.chains)
+
+    def contains(self, symbol_index: int, name: bytes) -> bool:
+        index = self.buckets[elf_sysv_hash(name) % len(self.buckets)]
+        while index:
+            if index == symbol_index:
```
JingsongLi: [P2] Make native symbol validation obey a linear work budget

This membership check restarts at the SysV bucket head for every exported symbol. A valid one-bucket table with the allowed 100,000 symbols requires roughly five billion chain reads; even a 5,000-symbol fixture performed 12,492,501 reads. Please retain the ownership or membership map built while validating the hash table so lookup is O(1). The same common symbol-count and string-scan budget should also be applied before the PE name_count and Mach-O symbol_count loops, which currently have no equivalent cap.

tools/verify_release_versions.py
jianguotian: Fixed. Both public entry points now resolve the root before comparing Cargo manifest paths.

tools/tests/test_validate_release_tag.py
jianguotian: Fixed. Lightweight-tag fixtures now explicitly disable tag.gpgSign.

tools/tests/test_validate_release_tag.py
jianguotian: Fixed. GPG fixtures now use a short canonical temporary root and clean it up.

tools/update_branch_version.sh
jianguotian: Fixed. POM updates now accept both OLD_VERSION and OLD_VERSION-SNAPSHOT, with regression coverage for the documented transition.

tools/native_binary.py
jianguotian: Fixed. Mach-O verification now counts only callable exports mapped to executable sections.

tools/verify_python_wheels.py
jianguotian: Fixed. All drive-qualified wheel paths, including C:../..., are now rejected.

tools/tests/test_validate_release_tag.py
jianguotian: Fixed. The revoked-key fixture now also uses the short temporary root.

tools/tests/test_verify_release_versions.py
jianguotian: Fixed. The fixture root is now resolved before relativizing updater results.

tools/verify_java_jars.py
jianguotian: Fixed. Every JAR member is now read through EOF so its CRC is validated.

tools/native_binary.py
jianguotian: Fixed. ELF hash ownership is built once for O(1) lookup, and PE/Mach-O symbol counts are capped.

PR #80, opened 2026-08-28
release: enforce signed-source preflight at publish boundaries

jianguotian:
## Summary

Gate release publication on a reusable signed-source preflight so release tags, component versions, and source artifacts are validated against the exact commit before credentialed publishing begins.

## Changes

- Add a reusable release preflight that, for tag runs, imports ASF `KEYS`, verifies an annotated signed tag bound to the current commit, and reads component versions from that signed Git tree.
- Create and verify deterministic source archives against raw Git blobs, with explicit size, entry, path, symlink, mode, and embedded-commit checks.
- Require the top-level release workflow and every credentialed Rust, Java, and Python publishing boundary to pass preflight; manual `workflow_dispatch` runs remain verification-only and cannot publish.
- Harden local source release creation with clean-tree and exact-tag checks, Git-environment isolation, GPG/SHA-512 verification, private staging, and no-overwrite behavior.
- Keep Rust package versions, Cargo path dependency constraints, `Cargo.lock`, Java, and Python versions aligned across documented branch transitions.
- Add the Release Vote Gate for `main` and `release-*`, plus adversarial regression tests for signature, source provenance, workflow dependencies, manual-dispatch guards, archive integrity, and version updates.
- Document RC retry behavior so prior local source artifacts are preserved before creating the next candidate.

## Validation

- [x] The branch contains exactly one commit on the current `apache/paimon-mosaic:main`.
- [x] `146 passed` across the focused release tooling test suite.
- [x] `actionlint 1.7.7` passed for all workflows.
- [x] Python compile checks, Bash syntax checks, YAML parsing, and `git diff --check` passed.
- [x] A source archive was created and independently verified against commit `01d8070aa2e5d01cc430043d8d04de837d6d5c7e`.
- [ ] Upstream GitHub Actions checks after PR creation.

## Notes

- This PR does not change the public API, runtime data format, or native read/write behavior.
- ASF vote outcome and final release authorization remain human governance steps; this PR supplies technical release gates only.
- Java JAR and Python wheel content verification remain outside this PR.
- Current review found no P0/P1 blockers. Symlinked Cargo path-dependency materialization and generic concurrent-output atomicity remain non-blocking P2 follow-ups rather than reasons to expand this release-gate PR.

### DIFF
--- a/.github/workflows/release-java.yml
+++ b/.github/workflows/release-java.yml
@@ -30,7 +30,11 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  release-preflight:
+    uses: ./.github/workflows/release-preflight.yml
+
   build-native:
+    needs: [release-preflight]
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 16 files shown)


JingsongLi:
+1

PR #83, opened 2026-09-06
fix(core): ignore NaN when tracking min/max column statistics

jackylee-ch:
## Purpose

`compare_values` returns `None` for any pair involving NaN, so a NaN bound can never be replaced. A FLOAT/DOUBLE row group whose first non-null value was NaN had both min and max pinned to NaN, making the range useless for row-group skipping.

## Changes

- Let a non-NaN value replace a NaN bound. An all-NaN column keeps NaN bounds, so a row group with any non-null value still has both set, as `serialize_stats` requires.

## Tests

- `test_stats_nan_does_not_pin_min_max`, `test_stats_all_nan_keeps_both_bounds_set`.


jianguotian:
For the 0.3.0 release, I propose treating this as a correctness blocker: NaN can pin min/max bounds and make row-group statistics unusable for pruning. The current head is mergeable and all five CI jobs are green.

PR #84, opened 2026-09-06
fix(core): reject empty column names when validating a schema

jackylee-ch:
## Purpose

`MosaicSchema::validate` rejects duplicate column names but not empty ones, while `deserialize` rejects them with "schema: empty column name". A writer therefore accepts an empty column name and closes successfully, producing a file no reader can open.

## Changes

- Reject empty column names in `validate`, matching the reader and the CLI check added in #81.

## Tests

- `test_validate_rejects_empty_column_name`, which also asserts such a schema still fails to deserialize.


jianguotian:
I reproduced one remaining 0.3.0 blocker: `MosaicWriter::from_mosaic_schema` accepts an already-built `MosaicSchema` and bypasses `MosaicSchema::validate`, so an empty column name can still be written and produce a file whose schema later fails to deserialize.

I opened a one-commit helper PR against this branch: https://github.com/jackylee-ch/paimon-mosaic/pull/1

Validation on the helper head: full `paimon-mosaic-core` tests passed, Clippy passed, fmt passed, and `git diff --check` passed. Please merge or cherry-pick it before this PR is included in 0.3.0.

PR #85, opened 2026-09-06
fix(core): return InvalidData instead of panicking on non-UTF-8 string columns

jackylee-ch:
## Purpose

`StringArray::new` is `try_new().unwrap()`, so reading a Utf8 column whose bytes are not valid UTF-8 panics inside arrow instead of returning an error. Across the JNI/FFI boundary a panic aborts the host process rather than raising.

## Changes

- Build the Utf8 array with `try_new` and map the failure to `InvalidData`. `build_array` already returns `io::Result`.

## Tests

- `test_build_array_non_utf8_returns_invalid_data_without_panicking`.


jianguotian:
For the 0.3.0 release, I propose treating this as a correctness blocker: malformed non-UTF-8 string data can otherwise panic instead of returning `InvalidData`. The current head is mergeable and all five CI jobs are green.

PR #87, opened 2026-09-10
feat(jni): add streaming columnar text JSON export

jianguotian:
## What

Add a Java API for streaming a prepared Mosaic row group as a JSON object whose values are comma-separated text columns, directly from Mosaic's native encodings and without materializing intermediate Arrow arrays.

## Example

Given a prepared Mosaic row group with these logical rows:

| Row | `timestamp` | `speed` | `state` |
| ---: | ----------: | ------: | ------- |
| 0 | 1 | 10 | `on` |
| 1 | 2 | `null` | `off` |
| 2 | 3 | 12 | `null` |

`ColumnarTextJsonWriter.write(rowGroup, output)` streams one JSON object:

```json
{"timestamp":"1,2,3","speed":"10,,12","state":"on,off,"}
```

Each JSON key is a projected column. Its string contains that column's values in logical row order, separated by commas. In other words, the three input rows above become one column-oriented text object rather than three row-oriented JSON objects.

This format is intentionally a text aggregation format, not a lossless typed serialization: `null` and an empty string both produce an empty entry, and commas inside string values remain literal.

The fast path changes the execution flow from:

```text
Mosaic encodings -> Arrow vectors -> Java serialization
```

to:

```text
Mosaic encodings -> native text writer -> OutputStream
```

The implementation:

- supports signed integers, `DOUBLE`, `Decimal128`, and UTF-8 values in `ALL_NULL`, `CONST`, `DICT`, and `PLAIN` encodings;
- preserves JVM-compatible `Double.toString` and `BigDecimal.toPlainString` output;
- returns `UNSUPPORTED` before touching the output when the row group requires fallback;
- keeps ownership of the caller's `OutputStream` with the caller;
- exposes a reusable `MosaicRowGroupReader` for fast-path attempts followed by Arrow fallback; and
- prevents concurrent or reentrant row-group operations while safely deferring native release when `close()` races an active operation.

The format, supported types, fallback behavior, and ownership rules are documented in `docs/columnar-text-json.md`.

## Why

Wide Mosaic files often contain many `CONST` and `ALL_NULL` columns. Reading those encodings directly avoids expanding every logical cell into Arrow vectors when a downstream interface only needs column-oriented text.

## Validation

Tests cover:

- primitive, dictionary, constant, all-null, and projected output;
- JVM-exact double formatting and decimal scales;
- unsupported and malformed encodings;
- partial-output and `OutputStream` exception behavior;
- output ownership; and
- reentrant use, concurrent close, idempotent close, and use-after-close lifecycle cases.

The source branch's identical tree passed the repository's `check`, `rust-test`, `cpp-test`, `java-test`, and `python-test` CI jobs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -921,6 +921,7 @@ dependencies = [
  "arrow-schema",
  "jni",
  "paimon-mosaic-core",
+ "ryu",
 ]
 
 [[package]]
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@
 
 A columnar-bucket hybrid format optimized for wide tables of Apache Paimon. 
 
+## Column-oriented text export
+
+The Java API can stream a row group as JSON containing comma-separated text columns,
+directly from Mosaic encodings without intermediate Arrow arrays.
+See the [format and API contract](docs/columnar-text-json.md).
+
 ## Java Linux compatibility
 
 Published Java artifacts include JNI libraries for Linux. The Linux GNU
@@ -48,4 +54,3 @@ Submit [issues](https://github.com/apache/paimon-mosaic/issues/new/choose) for b
 ## License
 
 Licensed under <a href="./LICENSE">Apache License, Version 2.0</a>.
-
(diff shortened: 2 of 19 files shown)


JingsongLi (approved):
+1

PR #88, opened 2026-09-11
reader: fetch small paged buckets whole when opening a row group

sundapeng:
## Purpose

With a partial projection, opening a row group needs two dependent round trips per paged bucket: the slot directory first, then the projected slots. On object storage each round trip costs tens of milliseconds while a small bucket is usually well under a megabyte, so read such buckets whole in round one and parse only the projected slots from that buffer. Buckets larger than `SMALL_PAGED_BUCKET_WHOLE_READ` (1 MiB) keep the directory-then-slots path.

## Changes

- `core/src/reader.rs`: paged buckets up to 1 MiB are fetched whole in round one even for partial projections; the whole-bucket branch now skips unprojected primary and child slots instead of assuming every column is projected.
- `core/src/reader_tests.rs`: a partially projected small paged bucket opens with exactly one read; the adjacent-slot coalescing test now expects the single whole-bucket read.

## Tests

`cargo test -p paimon-mosaic-core --release`: 158 core tests pass; `cargo clippy --tests -D warnings` and `cargo fmt` are clean.

- `test_paged_partial_projection_large_bucket_two_rounds` (new): a paged bucket above the threshold with a partial projection still takes the directory read followed by one coalesced slot read, and every decoded value is checked, so the two-round path keeps a value-checking test.


sundapeng:
cc @jianguotian — the Java side of this series is merged in apache/paimon (#9739, #9740, #9741, #9742, #9744); this is the remaining native piece.

PR #90, opened 2026-09-20
perf(core): build batch column map in linear time

jianguotian:
## Purpose

Build the Writer's sorted-column-to-input-batch mapping in linear time instead of repeatedly searching the Arrow schema during Writer construction.

## Changes

- invert `MosaicSchema::original_order` to construct the batch column map in O(N)
- remove repeated `Schema::index_of` calls from Writer construction
- preserve the existing sorted-column and input-batch mapping semantics
- add a focused test with a non-trivial column permutation

## Performance evidence

A local 1,000-column benchmark reduced Writer construction from 1,682 us to 218 us, approximately 87%. This result applies to Writer construction in that benchmark scenario, not to every batch write or whole-process execution.

## Compatibility

- no Mosaic file-format change
- column ordering and written values remain unchanged
- no FFI or public API change

## Tests

- `cargo deny check licenses`
- `git diff --check HEAD^`


JingsongLi:
+1

PR #95, opened 2026-09-21
test: make cross-language interop runs explicit

jianguotian:
## Purpose

The default Rust workspace test depended on Java and Python files under `/tmp/mosaic_interop`. A clean voter environment therefore failed, while a machine with stale fixtures could appear to pass without reproducing the cross-language producers.

## Changes

- mark the two cross-language Rust consumers as explicit ignored tests;
- keep the default Rust workspace test self-contained and visibly report the omitted interop consumers;
- run each consumer explicitly in CI after its Java or Python producer;
- document clean Java and Python interop sequences for release voters.

## Validation

- clean default workspace tests: 516 passed, 2 explicitly ignored
- missing-fixture explicit consumer: fails as expected
- Java interop: Rust writer 7/7, Maven 91/91, Rust reader 1/1
- Python interop: Rust writer 7/7, pytest 8/8, Rust reader 1/1
- workspace Clippy with `-D warnings`: passed
- `cargo fmt --all -- --check`: passed
- workflow YAML and `git diff --check`: passed


JingsongLi (approved):
Requirement fit: SUPPORTED. Implementation: CLEAN.

The default test behavior, explicit CI consumer invocations, and documented producer/consumer order are consistent. Verified `cargo test --locked --offline --workspace`: 514 passed, 2 ignored. Both explicit consumer commands run exactly one test and fail as expected when their corresponding fixture is missing. No actionable findings.

The complete Java/Python round trips were not rerun locally; official CI is still pending.

